PR #3, opened 2024-05-14
refactor: Update DAL to use SQLite3 instead of Badger

Pop0097:


cli/cmd/add.go
```diff
@@ -5,50 +5,46 @@ package cmd
...
-		command := Command{
-			Command: cmdText,
-			Tag:     tag,
-			Note:    note,
-			Alias:   alias,
-			Id:      nil,
+		// We must always have an alias. If one is not provided
+		// we will use the command text as the alias
+		if alias == "" {
+			alias = cmdText
```
danyal002, line 27: I don't think we need the alias anymore because we can use FTS (full text search) on SQL. It can search through entire objects and rows for us. We need an experiment for this and see how it works.

cli/cmd/add.go
```diff
@@ -5,50 +5,46 @@ package cmd
...
-		command_str, err := command.Serialize()
+		data_access_layer, err := dal.NewDataAccessLayer()
 		if err != nil {
 			log.Fatal(err)
+			return
 		}
+		defer data_access_layer.CloseDataAccessLayer()
 
-		err = dal.Add(strconv.FormatUint(*command.Id, 10), string(command_str))
+		err = data_access_layer.AddCommand(alias, cmdText, tags, note, 0)
```
danyal002: Did you hardcode this on purpose to come back and auto increment the primary key?

cli/cmd/list.go
```diff
@@ -13,13 +12,8 @@ import (
...
 	Use:   "list",
-	Short: "A brief description of your command",
-	Long: `A longer description that spans multiple lines and likely contains examples
-and usage of using your command. For example:
-
-Cobra is a CLI library for Go that empowers applications.
-This application is a tool to generate the needed files
-to quickly create a Cobra application.`,
+	Short: "List your most recently used commands in command stack",
+	Long:  `missing_docs`,
```
danyal002, line 16: I would do what we talked about yesterday. The main command would be to list all commands with a flag like `-r` to list recently used. You can also add `-l` for length to limit the number of rows. However, this last part can be done in another PR.

cli/dal/command.go
```diff
@@ -0,0 +1,35 @@
...
+	"encoding/json"
+)
+
+type Command struct {
+	Id       *uint64
+	Alias    string
+	Command  string
+	Tags     string
+	Note     string
+	UserId   *uint64
```
danyal002: I'd remove this field. We don't need it right now.

cli/dal/command.go
```diff
@@ -0,0 +1,35 @@
...
+	return json.Marshal(c)
+}
+
+func DeserializeCommand(data []byte) (Command, error) {
+	var cmd Command
+	err := json.Unmarshal(data, &cmd)
+	return cmd, err
+}
+
+func PrintCommands(commands []Command) {
```
danyal002, line 30: can this be a method too?

EDIT: I think this will get deleted in future iterations, so it doesn't matter atm.

cli/dal/dal.go
```diff
@@ -1,90 +1,86 @@
...
+	last_used := time.Now().Unix()
+	_, err := dal.db.Exec("INSERT INTO command (alias, command, tags, note, user_id, last_used) VALUES (?, ?, ?, ?, ?, ?)", alias, command, tags, note, user_id, last_used)
+	return err
 }
 
-func Add(key string, value string) error {
-	db, err := badger.Open(badger.DefaultOptions(DATABASE_DIRECTORY))
+// Search for a command by the given text
+func (dal *DataAccessLayer) SearchByCommand(command string) ([]Command, error) {
+	stmt, err := dal.db.Prepare("SELECT * FROM command WHERE command LIKE ?")
```
danyal002, line 63: I wonder if there's an option we need to turn on for FTS because this might be too slow. No need to fix this now.

cli/dal/parameter.go
```diff
@@ -0,0 +1,22 @@
+package dal
+
+import "encoding/json"
+
+type Parameter struct {
+	Id           *uint64
+	CommandId    *uint64
```
danyal002: what's the difference between these two?

danyal002 (changes requested):

cli/cmd/list.go
Pop0097: I plan on implementing this feature in another PR

cli/dal/parameter.go
Pop0097: Id is the unique identifier of the Parameter row in the `parameter` table. CommandId is a foreign key reference to the command that the parameter was created for. I can add a comment for that.

cli/cmd/add.go
Pop0097: We will most likely want the alias when we build the GUI

cli/cmd/add.go
Pop0097: The `0` is the user id. Since we don't have the concept of a user right now, I will remove it.

cli/dal/dal.go
Pop0097: I think I will save FTS for a separate PR since it is not essential at the moment.

danyal002 (approved):

PR #4, opened 2024-05-15
feat: allow users to list commands

Pop0097:
Defaults to listing 25 commands. use `--limit|-l` to change number of fields shown. Use `-r` to order results by most recently used.


cli/dal/dal.go
```diff
@@ -62,13 +63,47 @@ func (dal *DataAccessLayer) AddCommand(alias string, command string, tags string
...
+		log.Fatal("SearchByCommand: Failed to execute query: ", err)
+		return nil, err
+	}
+	defer rows.Close()
+
+	var commands []Command
+	for rows.Next() {
+		var command Command
+		if err := rows.Scan(&command.Id, &command.Alias, &command.Command, &command.Tags, &command.Note, &command.LastUsed); err != nil {
+			log.Fatal("SearchByCommand: Failed to extract Command: ", err)
```
danyal002: I think this is more of a serialization failure

cli/dal/dal.go
```diff
@@ -62,13 +63,47 @@ func (dal *DataAccessLayer) AddCommand(alias string, command string, tags string
...
+			return nil, err
+		}
+		commands = append(commands, command)
+	}
+	return commands, nil
+}
+
+// Get all commands from the database, limiting and/or ordering results based on the supplied parameters
+func (dal *DataAccessLayer) GetCommands(limit int, order_by_use bool) ([]Command, error) {
```
danyal002: I would use a different name for `order_by_use`. Something like `order_by_recent_usage`

danyal002 (changes requested):

danyal002 (approved):

PR #5, opened 2024-05-16
Delete commands by ID

Pop0097:
`go run main.go delete -i <ID>`

The user is presented with a success/error statement if:
1. The command does not exist in the database
2. The deletion was successful


cli/cmd/delete.go
```diff
@@ -15,20 +17,34 @@ var deleteCmd = &cobra.Command{
...
+		id, _ := cmd.Flags().GetInt("id")
+
+		if id == -1 {
+			log.Fatal("You must provide an ID to delete a command")
+			return
+		}
+
+		data_access_layer, err := dal.NewDataAccessLayer()
+		if err != nil {
+			log.Fatal(err)
```
danyal002: I would be consistent like your other PR and have a more descriptive error message.

cli/cmd/delete.go
```diff
@@ -15,20 +17,34 @@ var deleteCmd = &cobra.Command{
...
+			return
+		}
+		defer data_access_layer.CloseDataAccessLayer()
+
+		err = data_access_layer.DeleteCommandById(id)
+		if errors.Is(err, dal.MissingCommandError) {
```
danyal002, line 35: I like this way of error handling. We did the same in Rust at Haven. I would go back and either modify your other PR to do the same or do it in another PR.

It might be a good idea to separate logic from interfaces/error definitions into different files.

cli/dal/dal.go
```diff
@@ -83,3 +111,25 @@ func (dal *DataAccessLayer) SearchByCommand(command string) ([]Command, error) {
...
+	// Determine if the id exists
+	_, err := dal.GetCommandById(id)
+	if err != nil {
+		return err
+	}
+
+	// Delete the id
+	stmt, err := dal.db.Prepare("DELETE FROM command WHERE id = ?")
+	if err != nil {
+		log.Fatal(err)
```
danyal002: Same comment about descriptive error message here.

cli/dal/dal.go
```diff
@@ -83,3 +111,25 @@ func (dal *DataAccessLayer) SearchByCommand(command string) ([]Command, error) {
...
+		return err
+	}
+
+	_, err = stmt.Exec(id)
+	if err != nil {
+		log.Fatal(err)
```
danyal002: here as well.

cli/dal/dal.go
```diff
@@ -83,3 +111,25 @@ func (dal *DataAccessLayer) SearchByCommand(command string) ([]Command, error) {
 	}
 	return commands, nil
 }
+
+func (dal *DataAccessLayer) DeleteCommandById(id int) error {
+	// Determine if the id exists
+	_, err := dal.GetCommandById(id)
+	if err != nil {
+		return err
+	}
```
danyal002: You don’t need to do this. I think SQL will _not_ return an error message if you delete something that doesn't exist. We can do the same.

danyal002 (changes requested):

danyal002:
> The user is presented with a success/error statement if:
> 2. The deletion was successful

I don't think you need to display anything if the deletion was successful regardless of whether the command exists at the time of deletion or not. Most CLI tools just exit once you run a deletion command.

cli/dal/dal.go
Pop0097: I tested it and it doesn't

danyal002 (approved):

PR #6, opened 2024-05-26
feat: allow users to search for commands

Pop0097:
This feature allows users the ability to search for commands using command line arguments. 

- `-a` for alias
- `-t` for tag
- `-c` for command

We search and filter in the following order
1. Tag
2. Command
3. Alias

Once the user selects a command, it is added to their clipboard and the last used property of the command is updated in the database. 

https://github.com/danyal002/cmd-stack/assets/52176616/90475bd1-cd32-4128-a732-92aaa024e6f2


cli/cmd/delete.go
```diff
@@ -24,14 +24,14 @@ var deleteCmd = &cobra.Command{
...
 		}
 
-		data_access_layer, err := dal.NewDataAccessLayer()
+		dataAccessLayer, err := dal.NewDataAccessLayer()
 		if err != nil {
 			log.Fatal("Delete Cmd: Failed to create data access layer:", err)
 			return
 		}
-		defer data_access_layer.CloseDataAccessLayer()
+		defer dataAccessLayer.CloseDataAccessLayer()
```
danyal002, line 32: We should decide on a variable naming convention and stick to it. If we're going to use camel case, then we should change it everywhere in future PRs.

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
+
+	prompt := promptui.Prompt{
+		Label:     "Tag",
+		AllowEdit: true,
+		Default:   cur_tag,
+	}
+	tag, err := prompt.Run()
+	if err != nil {
+		log.Fatal("Search Cmd: Failed to prompt for tag", err)
+		return "", "", "", "", err
```
danyal002: Hmm, I don't like this. This introduces more edge cases. We should either have a `nil` returned or an actual value. What do you think?

(This is coming from me programming in Rust and getting annihilated for doing the above in my PRs)

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
+func init() {
+	rootCmd.AddCommand(searchCmd)
+	searchCmd.Flags().StringP("command", "c", "", "Search by command")
+	searchCmd.Flags().StringP("alias", "a", "", "Search by alias")
+	searchCmd.Flags().StringP("tag", "t", "", "Search by tag")
+	searchCmd.Flags().StringP("print", "p", "all", "Select how commands are presented to you (all, command, alias)")
+	searchCmd.Flags().IntP("limit", "l", 10, "Limit the number of commands to display before scrolling (defaults to 10)")
+}
+
+func searchArgsWizard(cur_tag string, cur_command string, cur_alias string) (string, string, string, string, error) {
```
danyal002: why the `cur_` prefix?

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
+	//	Label: "Select printing style",
+	//	Items: dal.CmdPrintingOptions,
+	//}
+	//_, printOption, err := sel.Run()
+	//if err != nil {
+	//	log.Fatal("Search Cmd: Failed to prompt for print option", err)
+	//	return "", "", "", "", err
+	//}
+
+	return tag, command, alias, printOption, nil
```
danyal002: You don't have to do this in this PR, but this should be a results object, something like `SearchResult`.

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
+}
+
+func extractAndValidateArgs(cmd *cobra.Command) (string, string, string, string, int, error) {
+	tag, _ := cmd.Flags().GetString("tag")
+	command, _ := cmd.Flags().GetString("command")
```
danyal002: Imo, the "command" or `-c` should be the default argument for the search command e.g. `cmdstack search "some command" -t ... -n ...` and so on.

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
+	alias, _ := cmd.Flags().GetString("alias")
+	printOption, _ := cmd.Flags().GetString("print")
+	limit, _ := cmd.Flags().GetInt("limit")
+
+	// If no arguments are provided, we will present the user with a form
```
danyal002: I think there's a better way to do this than what's below, but I can't remember off the top of my head. I remember reading somewhere that you can autogenerate this part using Cobra.

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
-			return
+			log.Fatal("Search Cmd: Failed to search for command by tag", err)
+			return nil, err
 		}
-		defer data_access_layer.CloseDataAccessLayer()
+	}
 
-		commands, err := data_access_layer.SearchByCommand(command)
+	if len(commands) > 0 && command != "" {
+		commands = dal.FilterCommandsByCommand(commands, command)
```
danyal002: Confusing name. A better name would be `FilterCommandsByCommandText` or `FilterCommandsByCommandContent`.

cli/cmd/search.go
```diff
@@ -5,42 +5,200 @@ package cmd
...
+		fmt.Println("No Commands Found...")
+		return
+	}
+
+	// Format the commands for printing
+	formattedCommands := dal.FormatCommands(commands, printOption)
+
+	// Prompt the user to select a command
+	prompt := promptui.Select{
+		Label: "Select Command (" + dal.GetPrintedValues(printOption) + ")",
```
danyal002: I'd remove the "Select Command" prefix. The little triangle on the left side of every command in the list is enough to know what the user is doing. It is also less cognitive load.

cli/dal/command.go
```diff
@@ -13,22 +15,96 @@ type Command struct {
...
+		tags = c.Tags + strings.Repeat(" ", MaxCharsPrintedTags-len(c.Tags))
+	}
+
+	if len(c.Note) > MaxCharsPrintedNote {
+		note = c.Note[:MaxCharsPrintedNote-3] + "..."
+	} else {
+		note = c.Note + strings.Repeat(" ", MaxCharsPrintedNote-len(c.Note))
+	}
+
+	return alias + " | " + command + " | " + tags + " | " + note
```
danyal002, line 54: Hmm, whilst this is fine for now, this will quickly become a maintenance issue because we'll be reusing it everywhere. I would suggest searching for a library that can take care of this for us. I'm sure one exists.

In Rust, we used something called `Validator` which would use attributes on each struct property to validate a struct every time it was created.

cli/dal/dal.go
```diff
@@ -87,32 +105,100 @@ func (dal *DataAccessLayer) GetCommandById(id int) (*Command, error) {
...
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal("SearchByCommand: Failed to create prepared statement:", err)
 		return nil, err
 	}
 
 	rows, err := stmt.Query("%" + command + "%")
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal("SearchByCommand: Failed to perform query:", err)
```
danyal002: nit: `Failed to execute query:`.

cli/dal/dal.go
danyal002, line 1: So, one big overarching comment I have is that I don't agree with your approach for search and filtering. A database is much better equipped to handle that for us. We're reinventing the wheel here. The way I would do this is by creating a `SearchFilters` object. You would then pass that to the `dal` `search` method and form your query using the filters. This is where `Squirrel` can help. You would validated the filters object there and return an error if the filters are invalid. Execute the query and create a `SearchResult` object to return.

danyal002 (changes requested):

cli/cmd/search.go
Pop0097: I agree, however with the way that the UI is generated by the library you need a prompt otherwise a `:` just appears.

cli/cmd/search.go
Pop0097: Can you? I had to use an external library for this

cli/dal/command.go
Pop0097: Will do in future PR

cli/cmd/search.go
Pop0097: Can address in future PR

cli/cmd/search.go
Pop0097: Will remove

cli/cmd/search.go
Pop0097: This is a good catch, thanks for the suggestion!

cli/cmd/delete.go
Pop0097: camel case

cli/cmd/list.go
```diff
@@ -16,27 +20,59 @@ var listCmd = &cobra.Command{
 	Long:  `missing_docs`,
 	Run: func(cmd *cobra.Command, args []string) {
 		use_recent, _ := cmd.Flags().GetBool("recent")
-		limit, _ := cmd.Flags().GetInt("limit")
+		resultLimit, _ := cmd.Flags().GetInt("result-limit")
+		printOption, _ := cmd.Flags().GetString("print")
+		printLimit, _ := cmd.Flags().GetInt("print-limit")
+		if !slices.Contains(dal.CmdPrintingOptions, printOption) {
+			fmt.Println("Invalid print argument")
+			log.Fatal("Search Cmd: Invalid print argument")
```
danyal002: Nit: could this be better named? I'm not sure what the difference between `resultLimit` and `printLimit` is.

cli/cmd/search.go
```diff
@@ -5,42 +5,150 @@ package cmd
...
-			return
-		}
-		defer data_access_layer.CloseDataAccessLayer()
+func init() {
+	rootCmd.AddCommand(searchCmd)
+	searchCmd.Flags().StringP("command", "c", "", "Search by command")
+	searchCmd.Flags().StringP("alias", "a", "", "Search by alias")
+	searchCmd.Flags().StringP("tag", "t", "", "Search by tag")
+	searchCmd.Flags().StringP("print", "p", "all", "Select how commands are presented to you (all, command, alias)")
+	searchCmd.Flags().IntP("limit", "l", 5, "Limit the number of commands to display before scrolling (defaults to 10)")
```
danyal002, line 28: We should talk about setting up a config file for all these settings.

cli/cmd/search.go
```diff
@@ -5,42 +5,150 @@ package cmd
...
+	if err != nil {
+		log.Fatal("Search Cmd: Failed to prompt for alias", err)
+		return nil, nil, nil, err
+	}
+
+	return &tag, &command, &alias, nil
+}
 
-		commands, err := data_access_layer.SearchByCommand(command)
+func extractAndValidateArgs(cmd *cobra.Command) (*dal.SearchFilters, *string, *int, error) {
```
danyal002: This would be better as a method I think, unless you plan on using it without a `Command`.

danyal002 (commented):
Could you move the `example_commands.txt` file to a separate `test_data` directory please?

cli/cmd/search.go
Pop0097: Do you mean a method of the DAL?

danyal002: the `Command` object as it's always called on a `Command`.

danyal002 (approved):
Approving with one comment

PR #7, opened 2024-05-26
feat: wizard for adding commands

Pop0097:
If the user does not add a value for `-a`, `-n`, or `-t`, then we will display a wizard


danyal002 (commented):
nit: I'm not a huge fan of storing data base artifcats in our github repo.

danyal002 (approved):

PR #8, opened 2024-06-01
feat: Add full-text-search

Pop0097:


Pop0097:
Searching doesn't quite work yet...

PR #12, opened 2024-06-23
Refactor: Switch to using Rust instead of Golang

Pop0097:


Pop0097:
Polishing that needs to be done in `cli`:
1. ~~When prompting users, it's a bit difficult to tell where the prompt ends and where their input begins:~~ Screenshot 2024-06-24 at 03 30 03
2. ~~I could probably share more code between delete, list, update, and search. Notably, the logic to determine if we should show the wizard or not.~~ 
3. ~~Both Add and Update declare their own versions of `set_command_properties_wizard`. Maybe I can figure out a way for them to share a function~~
4. ~~Need to make the printing of commands nicer-looking...~~ Screenshot 2024-06-24 at 03 33 09

Polishing that needs to be done in business logic:
1. ~~We currently create a new database connection on every call to the `logic` crate. If we can find a way so we only make one connection per cli invocation, that would be best.~~
2. ~~Searching is sometimes giving inaccurate results~~

cli/src/add_command.rs
```diff
@@ -0,0 +1,77 @@
+use crate::args::AddArgs;
+use data::models::InternalCommand;
+use inquire::{InquireError, Select, Text};
+
+#[derive(Debug)]
+/// The properties of a command
```
danyal002: BRO at least make an effort to show that this wasn't chatgpt generated 💀

cli/src/add_command.rs
```diff
@@ -0,0 +1,77 @@
...
+struct AddCommandProperties {
+    alias: String,
+    tag: Option<String>,
+    note: Option<String>,
+    favourite: bool,
+}
+
+/// Generates a wizard to set the properties of a command
+fn set_command_properties_wizard(command: &str) -> Result<AddCommandProperties, InquireError> {
```
danyal002: minor: do you know what `Inquire` errors look like? Idk if we should show them to the user or wrap them and dump it to a log. Regardless, you can fix this in another PR.

cli/src/add_command.rs
```diff
@@ -0,0 +1,77 @@
...
+    let mut tag = args.tag;
+    let mut note = args.note;
+    let mut favourite = args.favourite;
+
+    // If no alias, tag, or note is provided, generate a wizard to get them
+    if alias.is_none() && tag.is_none() && note.is_none() {
+        let command_properties = match set_command_properties_wizard(&command) {
+            Ok(properties) => properties,
+            Err(e) => {
+                println!("Add Cmd: Error setting command properties: {:?}", e);
```
danyal002: minor: make an error type for this using `thiserror`.

cli/src/add_command.rs
```diff
@@ -0,0 +1,77 @@
...
+        favourite = command_properties.favourite;
+    } else if alias.is_none() {
+        // If the alias is not provided, set it equal to the command
+        alias = Some(command.clone());
+    }
+
+    // Add the command to the database
+    let add_result = logic::command::handle_add_command(InternalCommand {
+        command: command,
+        alias: alias.unwrap(),
```
danyal002: I don't know what the solution is here and why I don't like this, but I don't. Generally, you should avoid unwrapping where possible.

Pop0097:
Polishing that could be done: 

1. **Logic Layer:**  Make parameter generation support more text encodings rather than just ASCII
2. ~~**Data Layer:** Make foreign key references work. For some reason ON DELETE CASCADE is not working when deleting a row in the command table~~
3. ~~**Data Layer:** Transactions~~
4. **CLI:** General UI improvements. I am not too happy about the fact we are showing a select prompt in the list commands. Also, I feel like the `param add` command could be improved. May also want to see if we want to implement multiselect when running `param update` and `param delete` so users can edit multiple parameters in one go rather than running the command multiple times. Also, probably most egregious is the fact we have `println!()` statements in places. If we can develop some sort of framework that will automatically manage spacing that would probably be ideal. 
5. **Quality of Life:** Logging and tests

danyal002 (approved):

PR #17, opened 2024-07-10
Add workflow to check rust

Pop0097:


Pop0097:
Failing because main uses go currently.

PR #23, opened 2024-10-03
Setup UI and list commands

Uzair5162:
Sets up the ui and displays the list of commands from the db. Run it with `yarn tauri dev` in `/ui`. Also excludes the ui from github actions because it breaks.


dylan-p-wong:
Why does ui break github actions?

Uzair5162:
> Why does ui break github actions?

https://github.com/danyal002/cmd-stack/actions/runs/11165028501/job/31035740593?pr=23#step:4:832

looks like some missing dependencies that we'd need to add to the workflow that I didn't bother trying to figure out yet

dylan-p-wong:
I think the build should pass. There should be a binary created for the desktop application.

dylan-p-wong:
I think we should do something like this https://v1.tauri.app/v1/guides/building/macos/

PR #24, opened 2024-10-08
Refactor and add tests

dylan-p-wong:
Refactor logic layer and data layer to allow use of temp dirs for testing.

Add tests for commands in logic layer.

### DIFF
--- a/cli/src/command/add_command.rs
+++ b/cli/src/command/add_command.rs
@@ -2,6 +2,7 @@ use crate::{args::AddArgs, command::print_internal_command, outputs::ErrorOutput
 use data::models::InternalCommand;
 use inquire::{InquireError, Select, Text};
 use log::error;
+use logic::new_logic;
 
 #[derive(Debug)]
 struct AddCommandProperties {
@@ -80,7 +81,14 @@ pub fn handle_add_command(args: AddArgs) {
         favourite,
     };
 
-    match logic::command::handle_add_command(internal_command.clone()) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Add Cmd", "Error adding command: {:?}", logic.err());
+        ErrorOutput::AddCmd.print();
+        return;
+    }
+
+    match logic.unwrap().handle_add_command(internal_command.clone()) {
         Ok(_) => {
             if !generate_command_with_wizard {
                 // If the user added the command via CLI arguments, we need to
--- a/cli/src/command/add_command.rs
+++ b/cli/src/command/add_command.rs
@@ -83,7 +83,7 @@ pub fn handle_add_command(args: AddArgs) {
 
     let logic = new_logic();
     if logic.is_err() {
-        error!(target: "Add Cmd", "Error adding command: {:?}", logic.err());
+        error!(target: "Add Cmd", "Failed to initialize logic: {:?}", logic.err());
         ErrorOutput::AddCmd.print();
         return;
     }
--- a/cli/src/command/add_command.rs
+++ b/cli/src/command/add_command.rs
@@ -88,7 +88,11 @@ pub fn handle_add_command(args: AddArgs) {
         return;
     }
 
-    match logic.unwrap().handle_add_command(internal_command.clone()) {
+    match logic
+        .as_ref()
+        .unwrap()
+        .handle_add_command(internal_command.clone())
+    {
         Ok(_) => {
             if !generate_command_with_wizard {
                 // If the user added the command via CLI arguments, we need to
--- a/cli/src/command/delete_command.rs
+++ b/cli/src/command/delete_command.rs
@@ -59,7 +59,7 @@ pub fn handle_delete_command(args: SearchAndPrintArgs) {
 
     let logic = new_logic();
     if logic.is_err() {
-        error!(target: "Delete Cmd", "Failed to delete command: {:?}", logic.err());
+        error!(target: "Delete Cmd", "Failed to initialize logic: {:?}", logic.err());
         ErrorOutput::FailedToCommand("delete".to_string()).print();
         return;
     }
--- a/cli/src/command/delete_command.rs
+++ b/cli/src/command/delete_command.rs
@@ -7,7 +7,7 @@ use crate::{
     outputs::ErrorOutput,
 };
 use log::error;
-use logic::command::SearchCommandArgs;
+use logic::{command::SearchCommandArgs, new_logic};
 
 /// UI handler for the delete command
 pub fn handle_delete_command(args: SearchAndPrintArgs) {
@@ -57,8 +57,15 @@ pub fn handle_delete_command(args: SearchAndPrintArgs) {
         },
     };
 
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Delete Cmd", "Failed to delete command: {:?}", logic.err());
+        ErrorOutput::FailedToCommand("delete".to_string()).print();
+        return;
+    }
+
     // Delete the selected command
-    match logic::command::handle_delete_command(selected_command.id) {
+    match logic.unwrap().handle_delete_command(selected_command.id) {
         Ok(_) => {}
         Err(e) => {
             error!(target: "Delete Cmd", "Failed to delete command: {:?}", e);
--- a/cli/src/command/delete_command.rs
+++ b/cli/src/command/delete_command.rs
@@ -65,7 +65,11 @@ pub fn handle_delete_command(args: SearchAndPrintArgs) {
     }
 
     // Delete the selected command
-    match logic.unwrap().handle_delete_command(selected_command.id) {
+    match logic
+        .as_ref()
+        .unwrap()
+        .handle_delete_command(selected_command.id)
+    {
         Ok(_) => {}
         Err(e) => {
             error!(target: "Delete Cmd", "Failed to delete command: {:?}", e);
--- a/cli/src/command/search_command.rs
+++ b/cli/src/command/search_command.rs
@@ -60,7 +60,7 @@ pub fn handle_search_commands(args: SearchAndPrintArgs) {
     let logic = new_logic();
     if logic.is_err() {
         error!(
-            target: "Search Cmd", "Failed to update command last used prop: {:?}",
+            target: "Search Cmd", "Failed to initialize logic: {:?}",
             logic.err()
         );
         ErrorOutput::GenerateParam.print();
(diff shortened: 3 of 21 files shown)


logic/src/lib.rs
```diff
@@ -6,15 +6,23 @@ pub mod command;
...
+impl Logic {
+    pub fn new(dal: SqliteDal) -> Logic {
+        Logic { db_connection: dal }
+    }
+}
+
+pub fn new_logic() -> Result<Logic, SQliteDatabaseConnectionError> {
+    let dal = SqliteDal::new()?;
+    Ok(Logic::new(dal))
+}
```
danyal002: This is not the best way to do this. We call or talk about it in person if you need more details. Essentially, we want an interface for the dal and abstract the SQLite stuff away. So, you want to implement a new interface called something like `Storage` or `StorageDal`. This will have some methods to `initialize()`, `add()`, `update()`, `delete()`, `get()` etc. These interface methods should take in objects like `InternalCommand` or `Command` or something like that. The database should be completely abstracted away. Once a command is read, these interface methods should be called. 

You will most likely need a conversion method to convert from `InternalCommand` to `Command` or vice-versa. Rust has great support for writing these [conversion methods.](https://doc.rust-lang.org/rust-by-example/conversion/from_into.html). 

In order to instrument our program for logs, we need to work on setting up the error handling right now. Once this interface is created, we need low-level, intermediate and user-level errors. This is easy to do with `this error`. Refer to `data/src/dal/sqlite.rs` to see how the conversion is happening.

logic/src/param.rs
```diff
@@ -59,99 +38,111 @@ pub enum GenerateParamError {
...
+    }
 
-    Ok(())
+    #[tokio::main]
+    pub async fn delete_param(&self, param_id: i64) -> Result<(), DefaultLogicError> {
+        // Delete the parameter from the database
+        match self.db_connection.delete_param(param_id, None).await {
+            Ok(_) => {}
+            Err(e) => return Err(DefaultLogicError::Query(e)),
+        };
```
danyal002, line 141: General comment for the `db_connection`. I think once the interface is implemented, we need to make a design decision on getting database connections. This important to think about because if our frontend is running at all times, we'll probably have a connection open on startup. We'll need to pool those connections. Let's talk in person.

danyal002 (changes requested):

logic/src/lib.rs
dylan-p-wong: Changed ```SqliteDal``` to ```Box<dyn Dal<Row = SqliteRow, DB = Sqlite>>```. I do not think there is a simple way to abstract SqlLite away completely if we are using transactions since a lot of the DAL methods contain ```async fn delete_command(
        &self,
        command_id: i64,
        tx: Option<&mut Transaction<'_, Self::DB>>``` where DB is some sqlx type.

cli/src/command/search_command.rs
```diff
@@ -57,7 +57,21 @@ pub fn handle_search_commands(args: SearchAndPrintArgs) {
         },
     };
 
-    let copied_text = match logic::param::handle_generate_param(selected_command.clone()) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(
+            target: "Search Cmd", "Failed to update command last used prop: {:?}",
+            logic.err()
+        );
```
Pop0097:
```suggestion
            target: "Search Cmd", "Failed to initialize logic: {:?}",
            logic.err()
        );
```

cli/src/command/add_command.rs
```diff
@@ -80,7 +81,18 @@ pub fn handle_add_command(args: AddArgs) {
         favourite,
     };
 
-    match logic::command::handle_add_command(internal_command.clone()) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Add Cmd", "Error adding command: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Add Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/command/delete_command.rs
```diff
@@ -57,8 +57,19 @@ pub fn handle_delete_command(args: SearchAndPrintArgs) {
         },
     };
 
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Delete Cmd", "Failed to delete command: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Delete Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/command/search_utils.rs
```diff
@@ -53,7 +53,16 @@ pub fn get_searched_commands(
     print_style: PrintStyle,
     display_limit: u32,
 ) -> Result<Command, GetSelectedItemFromUserError> {
-    let commands = match handle_search_command(search_args) {
+    let logic = match new_logic() {
+        Ok(l) => l,
+        Err(_) => {
+            return Err(GetSelectedItemFromUserError::GetCommands(
+                logic::DefaultLogicError::DbConnection(DatabaseConnectionError::InitDBConnection),
```
Pop0097: **NIT:** We may want a different error message for initializing the logic instead of returning the `InitDbConnetion` error.

cli/src/command/update_command.rs
```diff
@@ -121,8 +121,15 @@ pub fn handle_update_command(args: SearchAndPrintArgs) {
         }
     };
 
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Update Cmd", "Failed to update command: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Update Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/import_export.rs
```diff
@@ -1,13 +1,20 @@
...
 use std::path::Path;
 
 /// UI handler for export command
 pub fn handle_export_command(args: ImportExportArgs) {
     let file_path = Path::new(&args.file);
 
-    match logic::import_export::create_export_json(file_path) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Export Cmd", "Failed to export command: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Export Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/import_export.rs
```diff
@@ -23,7 +30,14 @@ pub fn handle_export_command(args: ImportExportArgs) {
 pub fn handle_import_command(args: ImportExportArgs) {
     let file_path = Path::new(&args.file);
 
-    match logic::import_export::import_data(file_path) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Import Cmd", "Failed to import command: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Import Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/param/add_param.rs
```diff
@@ -67,7 +68,14 @@ pub fn handle_add_param_command(command: Command) {
         }
     };
 
-    match logic::param::handle_add_param(params) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Add Param Cmd", "Error adding parameters: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Add Param Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/param/delete_param.rs
```diff
@@ -22,7 +23,14 @@ pub fn handle_delete_param_command(params: Vec<Parameter>, print_limit: u32) {
         },
     };
 
-    match logic::param::delete_param(param_to_delete.id) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Param Delete Cmd", "Error deleting parameter: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Param Delete Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/param/mod.rs
```diff
@@ -67,8 +68,15 @@ pub fn handle_param_command(param_command: ParamCommands) {
         },
     };
 
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Param Cmd", "Error getting parameters: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Param Cmd", "Failed to initialize logic: {:?}", logic.err());
```

cli/src/param/update_param.rs
```diff
@@ -67,7 +68,18 @@ pub fn handle_update_param_command(params: Vec<Parameter>, print_limit: u32) {
         }
     };
 
-    match logic::param::update_param(param_to_update.id, updated_internal_params) {
+    let logic = new_logic();
+    if logic.is_err() {
+        error!(target: "Param Update Cmd", "Error updating parameter: {:?}", logic.err());
```
Pop0097:
```suggestion
        error!(target: "Param Update Cmd", "Failed to initialize logic: {:?}", logic.err());
```

Pop0097 (approved):
Really great work, Dylan. I just had a few nits for our log messages.

logic/src/param.rs
Pop0097: I don't think we need to worry about this as Sqlite should provide thread safety automatically: https://sqlite.org/faq.html#q5

logic/src/lib.rs
Pop0097: I think the latest changes Dylan pushed up are more than sufficient. Since we don't plan on building out a backend anymore, we most likely will not use a different datastore than sqlite3. I'd say it's better to put our time towards building out the GUI and making the CLI more usable.

logic/src/lib.rs
```diff
@@ -36,32 +46,3 @@ pub enum DatabaseConnectionError {
     #[error("Got none after initializing db connection")]
     NoneAfterInit,
```
danyal002, line 55: Nit: this isn't your code, but a good time to fix this. The enum already tells me what the error is about, so the description isn't required or it can say something else like "Could not initialize database". When we instrument our application, the logs will make much more sense because the error message will immediately tell you the problem, and the stack trace will immediately tell you why it happened because of the enums.

danyal002 (approved):

cli/src/command/search_utils.rs
dylan-p-wong: Changed to ```return Err(GetSelectedItemFromUserError::GetCommands(e));```

logic/src/lib.rs
dylan-p-wong: Not addressing in this PR.

PR #25, opened 2024-10-28
setup tauri v2 and get command list from logic layer

Uzair5162:
Sets up the Tauri v2 ui and displays the list of commands from the db.
- Run it with `yarn tauri dev` in `/ui`.
- Excludes the Tauri project from the existing CI and adds a `test-tauri` workflow that builds the binary for all platforms (see https://v2.tauri.app/distribute/pipelines/github/)


dylan-p-wong (approved):
Very nice. It also seems fairly simple to setup the binary distribution by editing the actions. Can you get a photo of the UI. It does not have to look nice just show some commands. We also need it for the presentation. The GitHub action does take a while for Tauri. Maybe run it only on merge.

danyal002 (approved):

PR #26, opened 2024-11-03
Add basic GUI styling

dylan-p-wong:
image

How to use shadcn https://ui.shadcn.com/docs/components/accordion.

For example, if we want to add an accordion.

```
npx shadcn@latest add accordion
```

This will then create a new file in the ui/components directory. It is not a library.


ui/src-tauri/Cargo.toml
```diff
@@ -21,5 +21,6 @@ tauri-build = { version = "2", features = [] }
 tauri = { version = "2", features = [] }
 tauri-plugin-shell = "2"
 serde = { version = "1", features = ["derive"] }
+rustls-pemfile = { version = "2.2.0" }
```
dylan-p-wong, line 24: Was having issues running ```yarn tauri dev``` so had to add this.

ui/src-tauri/src/lib.rs
```diff
@@ -1,13 +1,19 @@
 use logic::new_logic;
+use serde::{Deserialize, Serialize};
 
-// Learn more about Tauri commands at https://tauri.app/develop/calling-rust/
-#[tauri::command]
-fn greet(name: &str) -> String {
-    format!("Hello, {}! You've been greeted from Rust!", name)
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
dylan-p-wong, line 5: Had to add this structure to translate the full command to the GUI. We can figure out a better solution for this later I just want to get the main frontend structure in main.

ui/src-tauri/src/lib.rs
```diff
@@ -17,19 +23,19 @@ fn list_commands() -> Result<Vec<String>, String> {
...
     };
-    let command_strings: Vec<String> = commands
+    let commands: Vec<Command> = commands
         .iter()
-        .map(|c| c.internal_command.alias.clone())
+        .map(|c| Command { id: c.id, last_used: c.last_used, alias: c.internal_command.alias.clone(), command: c.internal_command.command.clone(), tag: c.internal_command.tag.clone(), note: c.internal_command.note.clone(), favourite: c.internal_command.favourite})
         .collect();
 
-    Ok(command_strings)
+    Ok(commands)
```
danyal002: Could you implement a `From` trait here instead of mapping every property? https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

danyal002 (changes requested):
I'll leave the frontend review for @Uzair5162

ui/package-lock.json
Uzair5162, line 1: We're using yarn and have a `yarn.lock`, so we shouldn't have a `package-lock.json`. We should use yarn equivalents for any npm command that might've created this during setup.

Uzair5162 (changes requested):

ui/src-tauri/src/lib.rs
dylan-p-wong: Done.

ui/package-lock.json
dylan-p-wong: Done good catch.

Uzair5162 (approved):

ui/src-tauri/src/lib.rs
```diff
@@ -17,19 +38,16 @@ fn list_commands() -> Result<Vec<String>, String> {
         Ok(c) => c,
         Err(e) => return Err(format!("Error listing commands: {:?}", e)),
     };
-    let command_strings: Vec<String> = commands
-        .iter()
-        .map(|c| c.internal_command.alias.clone())
-        .collect();
+    let commands: Vec<DisplayCommand> = commands.iter().map(|c| DisplayCommand::from(c)).collect();
```
danyal002, line 41: minor minor nit: the rusty way of doing this is to `.map(|c| c.into()).collect();` (I think) because implementing `From` gives you the `Into` for free.  If you do `.map(|c| c.into()).collect();`, I have a feeling that cargo is going to tell you to do `.map(Into::into).collect();` instead which is even more idiomatic.

danyal002 (approved):

PR #27, opened 2024-11-13
Add parameters

dylan-p-wong:
Removes old parameter stuff.

Adds new parameters generation.

For example:
```
cmdstack add "git status ${int} ${string[0, 7]}"
```
When retrieved:
```
git status <random-int> <random-string-length-0-7>
```


logic/src/param.rs
```diff
@@ -1,145 +1,254 @@
...
+    fn default() -> Self {
+        Self { min: 5, max: 10 }
+    }
+}
+
+impl Parameter for StringParameter {
+    fn generate_random(&self, rng: &mut dyn RandomNumberGenerator) -> String {
+        let charset: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+        let length = rng.generate_range(self.min as i32, self.max as i32) as usize;
```
Uzair5162, line 72: this panics when `self.max <= self.min`. maybe we wanna check this and raise an error when parsing `StringParams` in: https://github.com/danyal002/cmd-stack/blob/92426a0ad115daddad254ff695d308e3e84fec9f/logic/src/param.rs#L87-L104

see `RandomNumberGenerator` comment as well since its related (https://github.com/danyal002/cmd-stack/pull/27#discussion_r1855723545).

logic/src/command.rs
```diff
@@ -222,6 +243,54 @@ mod tests {
...
+            .unwrap()
+            .path()
+            .to_string_lossy()
+            .into_owned();
+        let dal = SqliteDal::new_with_directory(path);
+        assert!(dal.is_ok());
+        let logic = Logic::new(Box::new(dal.unwrap()));
+
+        let mut invalid_command = InternalCommand {
+            command: "${bad}".to_string(),
```
Uzair5162: reminder to update this and below to `@{...}`

logic/src/param.rs
```diff
@@ -1,145 +1,254 @@
...
-    pub async fn handle_add_param(
-        &self,
-        params: Vec<InternalParameter>,
-    ) -> Result<(), AddParamError> {
-        for param in params.iter() {
-            if param.symbol.trim().is_empty() || param.regex.trim().is_empty() {
-                return Err(AddParamError::InvalidParam);
-            }
+impl RandomNumberGenerator for MockRng {
+    fn generate_range(&mut self, low: i32, high: i32) -> i32 {
```
Uzair5162, line 37: this impl is inconsistent with the `generate_range` impl for `ThreadRng`. `ThreadRng` calls `gen_range(low..high)` which generates a num in range `[low, high)`, while this ends up generating a num in range `[low, high]` (`MockRng` includes `high` and `ThreadRng` doesn't). This makes the test [here](https://github.com/danyal002/cmd-stack/blob/92426a0ad115daddad254ff695d308e3e84fec9f/logic/src/param.rs#L231) pass, even though it fails in use since the `ThreadRng` treats `[7,7]` as an empty range.

I'd lean towards keeping both bounds inclusive, so changing `ThreadRng` to call `gen_range(low..high+1)`.

Uzair5162 (changes requested):
good stuff

logic/src/param.rs
dylan-p-wong: Good catch. I changed to ```gen_range(low..high+1)``` and added test cases. I kept inclusive.

logic/src/param.rs
dylan-p-wong: Added.

Uzair5162 (approved):

danyal002 (approved):

PR #28, opened 2024-12-27
Add Form

dylan-p-wong:
image

Add form

Add Boolean Parameter

Renaming

### DIFF
--- a/cli/src/command/search_utils.rs
+++ b/cli/src/command/search_utils.rs
@@ -215,7 +215,7 @@ fn format_internal_commands(commands: &Vec<Command>) -> Vec<String> {
     }
 
     let table_str = table.to_string();
-    return table_str.lines().map(|s| s.to_string()).collect();
+    table_str.lines().map(|s| s.to_string()).collect()
 }
 
 pub fn copy_text(cmd: &str, text_to_copy: String) {
--- a/cli/src/command/search_utils.rs
+++ b/cli/src/command/search_utils.rs
@@ -3,7 +3,7 @@ use cli_clipboard::{ClipboardContext, ClipboardProvider};
 use data::models::Command;
 use inquire::{InquireError, Select, Text};
 use log::error;
-use logic::{command::SearchCommandArgs, new_logic};
+use logic::{command::SearchCommandArgs, Logic};
 use prettytable::{format, Cell, Row, Table};
 use termion::terminal_size;
 use thiserror::Error;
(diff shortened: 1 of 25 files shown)


cli/src/command/search_utils.rs
```diff
@@ -150,7 +150,7 @@ fn format_commands_for_printing(
     match print_style {
         PrintStyle::All => (
             format_internal_commands(commands),
-            "(Alias | Command | Tag | Note | Favourite [YES/NO])",
```
Pop0097: We don't display the ID in the search results. Are we planning on displaying the ID later? 
 
Screenshot 2024-12-30 at 15 30 29

cli/src/command/search_utils.rs
```diff
@@ -18,11 +18,11 @@ pub fn display_search_args_wizard(
 
 /// Generates a wizard to set the properties for command searching
 pub fn search_args_wizard() -> Result<SearchCommandArgs, InquireError> {
-    let command = Text::new("Command").prompt()?;
+    let command = Text::new("Command (Leave blank for no filter):").prompt()?;
```
Pop0097, line 21: This is a good enhancement. Nice work

ui/src/components/add-form.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }),
+  alias: z.string().min(1, {
+    message: 'Alias must be at least 1 character.',
+  }),
+  tag: z.string(),
+  note: z.string(),
+  favourite: z.boolean(),
+});
+
+export function AddForm() {
```
Pop0097: Would it be possible for the text boxes to resize relative to the width of the dialog? The gap between the text boxes and the right edge of the dialog looks a bit weird to me

Pop0097:
Another nit on my side: I think the UX would be a lot better if the dialog closes automatically on submission. 

https://github.com/user-attachments/assets/a1ab0b03-3b8f-4db5-b5f9-b350618df0c9

ui/src/components/add-form.tsx
dylan-p-wong: image

Changed. We still need to do quite a few UI fixes later. I mainly wanted to handle general logic in this PR.

cli/src/command/search_utils.rs
dylan-p-wong: I removed the ID. I think I was trying to do something different but forgot to change it back. We should not need to display the ID.

Pop0097 (approved):
Nice work

PR #31, opened 2025-01-11
Clean up DAL methods

Pop0097:
Updating the DAL methods to make them more idiomatic.

Also, I tried to make our CRUD methods more closely match the example in the sqlx repo: https://github.com/launchbadge/sqlx/blob/main/examples/sqlite/todos/src/main.rs

Same goes for transaction management: https://github.com/launchbadge/sqlx/blob/main/examples/postgres/transaction/src/main.rs


data/src/dal/mod.rs
```diff
@@ -5,17 +5,23 @@ use thiserror::Error;
 
 #[derive(Error, Debug)]
 pub enum SqlQueryError {
-    #[error("failed to add command")]
-    AddCommand(#[source] sqlx::Error),
+    #[error("failed to select commands")]
```
danyal002: NIT: Could you please capitalize the first letter of the error? ("Failed" instead of "failed") Just want to be consistent.

data/src/dal/sqlite_dal.rs
```diff
@@ -2,134 +2,88 @@ use super::sqlite::{SqliteConnectionPool, SqliteDbConnectionError};
...
-        directory: String,
+    pub async fn new_with_custom_path(
+        custom_path: String,
     ) -> Result<SqliteDal, SqliteDbConnectionError> {
-        let sqlite_db = SqliteConnectionPool::new(Some(directory)).await?;
-        Ok(SqliteDal { sql: sqlite_db })
+        let sqlite_db = SqliteConnectionPool::new(Some(custom_path)).await?;
+        Ok(SqliteDal {
+            sqlite_conn: sqlite_db,
+        })
```
danyal002, line 36: Since we're fixing the database path now, I think it's a good time to add an environment variable. An environment variable like `CMDSTACK_DB_PATH` is good.  The env var's existence is checked on start up. It is created and a default path is added to it if it doesn't exist. Everywhere else where we need the database path, it is fetched from this environment variable. This will clean up some code. You can make a method to fetch the environment variable and call it to get the default path. This will allow us to configure the default path depending on the OS type compilation very easily later down the road.

sqlx also does this based on the url you provided in the description. `let pool = SqlitePool::connect(&env::var("DATABASE_URL")?).await?;`

data/src/dal/sqlite_dal.rs
```diff
@@ -2,134 +2,88 @@ use super::sqlite::{SqliteConnectionPool, SqliteDbConnectionError};
...
-        if let Some(trans) = tx {
-            sqlx::query(query).execute(&mut **trans).await.map(|_| ())
-        } else {
-            sqlx::query(query).execute(&self.sql.pool).await.map(|_| ())
-        }
+    ) -> Result<SqliteQueryResult, sqlx::Error> {
+        let result = match tx {
+            Some(transaction) => sqlx::query(query).execute(&mut **transaction).await?,
+            None => sqlx::query(query).execute(&self.sqlite_conn.pool).await?,
+        };
```
danyal002: NIT: Usually recommended to _not_ use the `match` statement for an if/else case.
```suggestion
    let result = if let Some(transaction) = tx {
        sqlx::query(query).execute(&mut **transaction).await?
    } else {
        sqlx::query(query).execute(&self.sqlite_conn.pool).await?
    };
```

data/src/dal/sqlite_dal.rs
```diff
@@ -2,134 +2,88 @@ use super::sqlite::{SqliteConnectionPool, SqliteDbConnectionError};
...
-        let rows;
-        if let Some(trans) = tx {
-            rows = sqlx::query(query).fetch_all(&mut **trans).await?;
-        } else {
-            rows = sqlx::query(query).fetch_all(&self.sql.pool).await?;
-        }
+        let rows = match tx {
+            Some(transaction) => sqlx::query(query).fetch_all(&mut **transaction).await?,
+            None => sqlx::query(query).fetch_all(&self.sqlite_conn.pool).await?,
+        };
```
danyal002: Same comment as above

data/src/dal/sqlite_dal.rs
```diff
@@ -165,10 +119,10 @@ impl SqliteDal {
...
-            Ok(rows) => rows,
-            Err(e) => return Err(SqlQueryError::SearchCommand(e)),
-        };
+        let rows = self
+            .read_rows(&query, tx)
+            .await
+            .map_err(|e| SqlQueryError::SelectCommand(e))?;
 
         let mut commands = Vec::new();
         for row in rows {
```
danyal002: Not your code, but it is more idiomatic in Rust to use `into_iter()` and `collect()`. It avoids making the variable `mut` which is always desirable. 

```
let commands: Vec<Command> = rows
    .into_iter()
    .map(|row| Command {
        internal_command: InternalCommand {
            alias: row.get("alias"),
            command: row.get("command"),
            tag: row.get("tag"),
            note: row.get("note"),
            favourite: row.get("favourite"),
        },
        id: row.get("id"),
        last_used: row.get("last_used"),
    })
    .collect();
```

data/src/dal/sqlite_dal.rs
```diff
@@ -188,23 +142,63 @@ impl SqliteDal {
         Ok(commands)
     }
 
+    /// Returns the ID of the inserted command
```
danyal002: NIT: a more helpful comment would be "Adds a new command and returns the `ID` of the inserted command"

data/src/dal/sqlite_dal.rs
```diff
@@ -188,23 +142,63 @@ impl SqliteDal {
...
+                command.note.into(),
+                command.favourite.into(),
+                current_time.into(),
+            ])
+            .to_string(SqliteQueryBuilder);
+
+        let result = self
+            .execute_query(&query, tx)
+            .await
+            .map_err(|e| SqlQueryError::InsertCommand(e))?;
```
danyal002: You can do
```suggestion
           .map_err(SqlQueryError::InsertCommand)?;
```
which does the same thing but slightly more clean.

data/src/dal/sqlite_dal.rs
```diff
@@ -188,23 +142,63 @@ impl SqliteDal {
...
             .to_string(SqliteQueryBuilder);
 
-        match self.execute(&query, tx).await {
-            Ok(_) => {}
-            Err(e) => return Err(SqlQueryError::UpdateCommandLastUsed(e)),
-        };
+        let result = self
+            .execute_query(&query, tx)
+            .await
+            .map_err(|e| SqlQueryError::UpdateCommand(e))?;
```
danyal002: Same comment as above

data/src/dal/sqlite_dal.rs
```diff
@@ -219,10 +213,9 @@ impl SqliteDal {
             .and_where(Expr::col(sqlite::Command::Id).eq(command_id))
             .to_string(SqliteQueryBuilder);
 
-        match self.execute(&query, tx).await {
-            Ok(_) => {}
-            Err(e) => return Err(SqlQueryError::UpdateCommandLastUsed(e)),
-        };
+        self.execute_query(&query, tx)
+            .await
+            .map_err(|e| SqlQueryError::DeleteCommand(e))?;
```
danyal002: Same comment as above

data/src/dal/sqlite_dal.rs
```diff
@@ -248,10 +241,14 @@ impl SqliteDal {
...
             .to_string(SqliteQueryBuilder);
 
-        match self.execute(&query, tx).await {
-            Ok(_) => {}
-            Err(e) => return Err(SqlQueryError::UpdateCommandLastUsed(e)),
-        };
+        let result = self
+            .execute_query(&query, tx)
+            .await
+            .map_err(|e| SqlQueryError::UpdateCommand(e))?;
```
danyal002: Same comment as above

danyal002 (changes requested):

data/src/dal/sqlite_dal.rs
dylan-p-wong: The main reason I did not add ```CMDSTACK_DB_PATH``` in the dal layer was so I could configure each test in the logic layer with a temp directory. Since tests run in parallel each would set ```CMDSTACK_DB_PATH``` and there would be race issues on the temp directory.

danyal002: That is not a good enough reason to not have `CMDSTACK_DB_PATH`. Core functionality always comes first and shouldn't _not_ be implemented because it makes testing somewhat harder. Tests can be modified to run in series or the env var can be tested in another way or startup code can be avoided in the tests. 

For example, you can always mock the method that returns the environment variable in your test code. This way, no environment variable is actually getting set.

dylan-p-wong: If a user needs to set ```CMDSTACK_DB_PATH``` I think they should do it in the logic layer.

Passing the desired directory as a parameter is essentially just mocking it. I see no reason to tie anything to ```CMDSTACK_DB_PATH``` at the dal layer. It seems like a very common pattern that data storage layers file storage is configured through a directory parameter (https://github.com/google/leveldb/blob/main/doc/index.md).

Summary: If we want ```CMDSTACK_DB_PATH``` implement this in the logic crate.

Pop0097: I can look into this in a future PR

danyal002: > The main reason I did not add CMDSTACK_DB_PATH in the dal layer was so I could configure each test in the logic layer with a temp directory. Since tests run in parallel each would set CMDSTACK_DB_PATH and there would be race issues on the temp directory.

@dylan-p-wong 
Sorry, I think I misread your previous comment. I agree with putting it in the logic layer. I thought you were saying that we can't have `CMDSTACK_DB_PATH` at all because it would break tests.

danyal002: > I can look into this in a future PR

@Pop0097 Could you please make a Jira for it so that we don't forget?

danyal002 (approved):

PR #34, opened 2025-01-12
[DRAFT] CLI Improvements

Pop0097:
The following changes are made:
1. Update the help docs
2. Add colours + more consistent spacing 

Here are some examples with different VS Code themes: 

Screenshot 2025-01-12 at 12 43 17
Screenshot 2025-01-12 at 12 43 23


cli/src/args.rs
```diff
@@ -9,96 +9,100 @@ pub struct CmdStackArgs {
...
-    /// Search for a command
+    /// Search for a command in your stack
     Search(SearchAndPrintArgs),
 
-    /// Export commands to a file
+    /// Export commands to a JSON file
     Export(ImportExportArgs),
 
-    /// Import commands from a file
+    /// Import commands from a JSON file
```
danyal002, line 27: "Import stack from a JSON file"

cli/src/args.rs
```diff
@@ -9,96 +9,100 @@ pub struct CmdStackArgs {
...
     #[clap(long = "alias", short = 'a')]
     pub alias: Option<String>,
 
-    /// The command description
+    /// Notes relating to the command (optional)
     #[clap(long = "note", short = 'n')]
     pub note: Option<String>,
 
-    /// The command tag
+    /// The tag for the command (optional)
```
danyal002, line 45: Is there a better way to show that this is optional by using `clap` instead of writing it in a string?

cli/src/args.rs
```diff
@@ -9,96 +9,100 @@ pub struct CmdStackArgs {
...
     #[clap(long = "tag", short = 't')]
     pub tag: Option<String>,
 
-    /// Add the command to your favourites
+    /// If set, favourites the command
```
danyal002, line 49: "Mark the command as favourite"

cli/src/outputs.rs
```diff
@@ -1,44 +1,180 @@
...
+    table.add_row(Row::new(vec![
+        Cell::new("Favourite:").with_style(Attr::Bold),
+        Cell::new(favourite_status),
+    ]));
+
+    table.printstd();
+}
+
+/// Printing vertical space
+pub fn spacing() {
```
danyal002, line 78: `const String space = "\n"`

PR #35, opened 2025-01-13
UI improvements

Uzair5162:
A bunch of UI improvements and bug fixes:
- fixes the window having a blank space on the right and bottom when large
- adds a larger default window size
- adds minimum window dimensions
- updates the window name and app icons to be the CMDStack logo
  - the name on the dock is still 'ui', haven't figured out how to change this yet
- fixes issue where the window as a whole scrolls for no reason, only the cmd list should scroll now (when exceeding window height)
- removes the account switcher in the top right corner and replaces it with the logo/name
- fixes the favourite button in the middle pane not actually showing favourites
- stubs a favourite button on the left panel (doesn't do anything)
- makes the primary colour blue (only used as button backgrounds as of now)

image


ui/src-tauri/tauri.conf.json
```diff
@@ -12,9 +12,11 @@
...
     "windows": [
       {
-        "title": "ui",
-        "width": 800,
-        "height": 600
+        "title": "CMDStack",
+        "width": 1000,
+        "height": 600,
+        "minWidth": 800,
+        "minHeight": 600
```
Pop0097: **Nit:** I think the min width is still a bit too small since the "ALL | FAVOURITES" button gets cut off

Screenshot 2025-01-12 at 23 46 32

ui/src/App.tsx
```diff
@@ -1,9 +1,9 @@
-import './index.css'
+import './index.css';
 import CommandPage from './page';
 
 function App() {
   return (
-    <main className="container">
+    <main>
```
Pop0097, line 6: **Nit:** It appears that if the height of the app is increased a lot, the content cuts off

Screenshot 2025-01-12 at 23 49 42

ui/src/components/command-list.tsx
```diff
@@ -1,28 +1,28 @@
...
+  items: Command[];
 }
 
 export function CommandList({ items }: CommandListProps) {
-  const [command, setCommand] = useCommand()
+  const [command, setCommand] = useCommand();
 
   return (
-    <ScrollArea className="h-screen">
+    <ScrollArea className="h-[calc(100vh-121px)]">
```
Pop0097, line 18: **Question:** Is 121px is the height of the header + search bar?

ui/src/components/command.tsx
```diff
@@ -67,39 +55,54 @@ export function MainCommandPage({
...
-                variant: "default",
+                variant: 'default',
+              },
+              {
+                title: 'Favorites',
+                label: commands
+                  .filter((item) => item.favourite)
+                  .length.toString(),
+                icon: Star,
+                variant: 'ghost',
```
Pop0097, line 105: **Question:** I'm guessing this would not be implemented in this PR, but I am curious to know how this would eventually support users selecting tabs.

https://github.com/user-attachments/assets/59ab23a1-13f4-4549-81f7-6449b58fe0e4

Pop0097 (commented):
Really amazing work, Uzair! The app looks so much better!

ui/src-tauri/tauri.conf.json
Uzair5162: Good catch, increased it to 860px.

ui/src/App.tsx
Uzair5162: Fixed.

ui/src/components/command-list.tsx
Uzair5162: Yes, slightly hacky but nothing else was working.

ui/src/components/command.tsx
Uzair5162: Assuming you meant tags, my thinking was if users wanted to filter by favourite on a certain tag, they would use the favourite button in the middle pane (All | **Favourites**) instead of this, and this could be used only when they want to see all favourited commands. We can discuss this more in our next meeting.

dylan-p-wong (approved):

ui/src-tauri/tauri.conf.json
Pop0097: Screenshot 2025-01-13 at 11 34 17

Hmm, this did not seem to fix the issue. Playing with the app more, it appears that the minimum width of the `Commands` column is a percentage of the width of the whole window. I think we may need to hard code a value for the minimum width for this column.

Pop0097: The same applies to the minimum width of the navigation column too.

danyal002:
Hmm, the logo doesn't seem right. Seems like the font doesn't render properly. Can you also get rid of the padding/margin around the logo? I think it'd look better if it was close to the boundary and slightly bigger. 

Can we also swap the positions of the search bar with the "Commands"? The search can go at the top and then stretch across.

Can we also change the title to `CmdStack` instead please?

ui/src-tauri/tauri.conf.json
Uzair5162: Okay I added an absolute minWidth to the middle panel and made the minWidth on the window 800px again. Gonna leave off the navigation column until its contents are more finalized (tags, btns, etc.)

Uzair5162:
> Hmm, the logo doesn't seem right. Seems like the font doesn't render properly. Can you also get rid of the padding/margin around the logo? I think it'd look better if it was close to the boundary and slightly bigger.
> 
> Can we also swap the positions of the search bar with the "Commands"? The search can go at the top and then stretch across.
> 
> Can we also change the title to `CmdStack` instead please?

Renamed. Lets do the Search Bar and logo reworks in future PRs.

Pop0097 (approved):

PR #36, opened 2025-01-14
Backend Refactor

Pop0097:


Pop0097:
TODO:
1. ~~Make the batch add method in the SQLite dal and remove all transaction logic (@danyal002)~~
2. ~~Fix the tests (that are inevitably broken)~~
3. ~~Fix FE references~~
4. Ensure printing in the CLI works. Get inspiration from: https://github.com/danyal002/cmd-stack/pull/34 (@Pop0097)

Danyal here:
So, I included an example of how error strings are being changed and bubbled up. I didn't do this for all possible errors - just the `import` command. The reason for that is I need to do some thinking about how much detail we're gonna show the users. In my previous job, we abstracted away the implementation error from the user, but we logged the entire chain. I'm thinking of doing the same here. So, next step will be implement logging.

.DS_Store
dylan-p-wong, line 1: Should ignore this file.

logic/src/import_export.rs
```diff
@@ -1,156 +1,110 @@
...
-    #[test]
-    fn is_file_json_no_extension() {
-        let dir = tempdir().unwrap();
-        let file_path = dir.path().join("test");
-
-        let result = is_file_json(&file_path);
-        assert!(matches!(result, Err(ImportExportError::InvalidFilePath)));
-    }
-}
+// #[cfg(test)]
```
dylan-p-wong: Why are these tests being removed?

logic/src/import_export.rs
```diff
@@ -1,156 +1,110 @@
...
-                Err(e) => {
-                    self.db_connection.rollback(tx).await?;
-                    return Err(ImportExportError::DbQuery(e));
-                }
-            };
-            import_cmd_id_to_db_id.insert(command.id, db_id);
-        }
-
-        self.db_connection.commit(tx).await?;
+    pub async fn import_data(&self, import_file_path: &Path) -> Result<i64, ImportError> {
```
dylan-p-wong: Let's add a simple test case for this. Or add a ticket on Jira so we don't forget.

data/src/dal/sqlite_dal.rs
```diff
@@ -174,16 +174,80 @@ impl SqliteDal {
...
 
         Ok(result.last_insert_rowid())
     }
 
-    pub async fn update_command_last_used_prop(
+    pub async fn insert_mulitple_commands(
+        &self,
+        commands: Vec<InternalCommand>,
+        tx: Option<&mut Transaction<'_, Sqlite>>,
+    ) -> Result<i64, InsertCommandError> {
```
dylan-p-wong: Not sure what is going on here.

logic/src/command.rs
```diff
@@ -124,89 +142,71 @@ impl Logic {
...
-            .db_connection
+        let commands = self
+            .dal
             .get_all_commands(order_by_use, favourite, None)
-            .await
-        {
-            Ok(results) => results,
-            Err(e) => return Err(DefaultLogicError::Query(e)),
-        };
+            .await?;
```
dylan-p-wong: Do we want a ```.map_err(ListCommandError::Database);``` here?

logic/src/command.rs
```diff
@@ -1,79 +1,97 @@
...
+    Database(#[from] SelectAllCommandsError),
+}
 
-    #[error("error generating params")]
-    GenerateParamError(#[from] ParameterError),
+#[derive(Error, Debug)]
+pub enum DeleteCommandError {
+    #[error("Invalid user input")]
+    InvalidInput,
+    #[error("Failed to add command")]
```
dylan-p-wong: Typo: add -> delete. Is there any invalid input for delete if it is just by id? Should there be something like NotFound.

data/src/dal/mod.rs
```diff
@@ -1,27 +1,39 @@
...
-
-    #[error("Failed to delete a command")]
-    DeleteCommand(#[source] sqlx::Error),
-
-    #[error("Failed to update a command")]
-    UpdateCommand(#[source] sqlx::Error),
+pub enum InsertCommandError {
+    #[error("Failed to get the unix timestamp")]
+    UnixTimestamp(#[from] std::time::SystemTimeError),
+    #[error("Expected rows to be affected by the operation but none were affected")]
```
dylan-p-wong, line 9: In what case does NoRowsAffected happen for insert? I feel there is a better error message for those cases?

cli/src/main.rs
```diff
@@ -92,17 +92,39 @@ fn main() {
     let args = CmdStackArgs::parse();
 
     match args.command {
-        Command::Add(add_args) => command::add_command::handle_add_command(add_args),
-        Command::Update(update_args) => command::update_command::handle_update_command(update_args),
-        Command::Delete(delete_args) => command::delete_command::handle_delete_command(delete_args),
+        Command::Add(add_args) => match command::add_command::handle_add_command(add_args) {
```
dylan-p-wong, line 96: I feel this is a weird pattern. Not sure if there is a better way to do this.

cli/src/command/update_command.rs
```diff
@@ -57,99 +71,37 @@ pub fn set_command_properties_wizard(
 }
 
 /// UI handler for the update command
-pub fn handle_update_command(args: SearchAndPrintArgs) {
```
dylan-p-wong, line 60: Nice this is a lot cleaner now.

cli/src/command/add_command.rs
```diff
@@ -27,85 +42,38 @@ fn set_command_properties_wizard(command: &str) -> Result<AddCommandProperties,
...
-        tag = command_properties.tag;
-        note = command_properties.note;
-        favourite = command_properties.favourite;
-    } else if alias.is_none() {
-        // If the alias is not provided, set it equal to the command
-        alias = Some(command.clone());
-    }
+/// CLI handler for the add command
+pub fn handle_add_command(args: AddArgs) -> Result<(), HandleAddError> {
+    let add_args_exist = args.alias.is_some() || args.tag.is_some() || args.note.is_some();
```
dylan-p-wong, line 69: I think there is a helper function for something like this in search.

dylan-p-wong: Which fields are required in terms of the logic layer? So that UI and CLI restrict the same things.

logic/src/import_export.rs
Pop0097: The function that they were testing (`is_json_file`) was removed

logic/src/import_export.rs
Pop0097: Created a ticket for this: https://cmdstack.atlassian.net/browse/KAN-20?atlOrigin=eyJpIjoiNzIxYTNlNWUyZGRjNDU1NTkwYjU2NTE2MGQwNDE4MmQiLCJwIjoiaiJ9

data/src/dal/sqlite_dal.rs
Pop0097: This method is still WIP. The goal is that this method will do a batch insert so we can remove all logic relating to transactions from our codebase. 

@danyal002 will implement it shortly.

logic/src/command.rs
Pop0097:
```
#[derive(Error, Debug)]
pub enum ListCommandError {
    #[error("Invalid user input")]
    InvalidInput,
    #[error("Failed to list command")]
    Database(#[from] SelectAllCommandsError),
}
```

Because of the `#[from]` tag, we don't need to explicitly write the `map_err` since it's handled automatically.

logic/src/command.rs
Pop0097: Correct, there is a possibility for invalid input if a command with the provided ID does not exist. If this is the case, the DAL will return `Err(data::dal::DeleteCommandError::NoRowsAffected)`. 

This is captured by `Err(logic::command::DeleteCommandError::Database)` in the logic crate. This will then be captured in the CLI crate and an error will be displayed to the user (future PR)

cli/src/command/add_command.rs
Pop0097: @dylan-p-wong, so far only the command and alias are required. However, we will phase out the alias soon, so only the command field will be required.

logic/src/command.rs
dylan-p-wong: I see most the other methods in this file using map_err. Can those be removed? Example update_command_last_used_prop.

Pop0097: Of course! Sorry some must've slipped through our checks 😅

cli/src/main.rs
danyal002: ~~That doesn't look right. It'll get fixed in a future PR or I'll fix it soon.~~
Sorry, what do you mean? What makes this pattern weird?

cli/src/command/search_utils.rs
```diff
@@ -192,20 +214,22 @@ fn format_internal_commands(commands: &Vec<Command>) -> Vec<String> {
...
+#[derive(Error, Debug)]
+pub enum CopyTextError {
+    #[error("Failed to initialize the clipboard")]
+    ClipboardInit,
+
+    #[error("Failed to copy text to clipboard")]
+    Copy,
+}
+
+pub fn copy_text(cmd: &str, text_to_copy: String) -> Result<(), CopyTextError> {
```
danyal002: @Pop0097 The `cmd` variable is no longer being used.

danyal002 (approved):

PR #37, opened 2025-01-14
Improve CLI Output

Pop0097:
# Success Cases

Screenshot 2025-01-14 at 18 15 19
Screenshot 2025-01-14 at 18 15 26
Screenshot 2025-01-14 at 18 15 48

# Errors
Screenshot 2025-01-14 at 18 16 56
Screenshot 2025-01-14 at 18 17 01


danyal002:
2 comments:
* Can we replace the `!` with a period? `.`
* I wonder what this would look like if we kept the text colour the same, but introduced an emoji at the front. So it would be like:
`✅ Command copied to clipboard.`
`❌ Failed to get input.`

It's very minor, but it's something I learned from one of my coops. They were talking about how humans read from left to right, and a checkmark or an X mark immediately conveys the message we're trying to pass if the emoji is in front of the success/error message. If you look at Windows error messages, they also follow the same doctrine:

image

PR #42, opened 2025-01-15
Flag commands that only contain whitespace

Pop0097:
Our previous validator would let commands with only whitespace pass (ex. `     `). I had to make a custom validator to ensure we trim the string before checking the length.

### DIFF
--- a/cli/src/utils.rs
+++ b/cli/src/utils.rs
@@ -10,8 +10,8 @@ pub fn truncate_string(s: &str, width: usize) -> String {
     }
 }
 
-/// Returns None if the provided string is empty. 
-/// If a string only contains whitespace, None is returned. 
+/// Returns None if the provided string is empty.
+/// If a string only contains whitespace, None is returned.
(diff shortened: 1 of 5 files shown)


cli/src/utils.rs
```diff
@@ -10,6 +10,16 @@ pub fn truncate_string(s: &str, width: usize) -> String {
...
+/// Returns None if the provided string is empty.
+/// If a string only contains whitespace, None is returned.
+pub fn none_if_empty(s: String) -> Option<String> {
+    if !s.trim().is_empty() {
+        Some(s)
+    } else {
+        None
+    }
+}
+
```
danyal002, line 22: You might be able to replace this method with just a filter:

`Some("some string").filter(|s| !s.trim().is_empty());` and that returns the string or `None`

danyal002 (approved):

cli/src/utils.rs
Pop0097: I thought of doing that, but repeating that code everywhere felt less clean than having a dedicated function

PR #43, opened 2025-01-16
Add UI tag directories

Uzair5162:
Adds a tree view component and a command tag directory to the sidebar.
- the tag directory is generated by parsing all of the tag strings into a tree
  - ex. `git/push/main`, `git/push/branch`, and `git/pull` get parsed into:
```
git
└───push
│   └───main
│   └───branch
└───pull
```
- clicking a tag filters the cmd list to only show commands with tags beginning with the selected tag
- the selected tag is displayed as a badge above the cmd list
- clicking the "All Tags" button above the tag tree clears the selected tag and shows all commands
- tag tree scrolls if overflow

image


ui/src/components/tag-tree.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const root: Record<string, TagNode> = {};
+
+    commands.forEach((command) => {
+      if (!command.tag) return;
+      const tagPath = command.tag.split('/');
+      let current = root;
+
+      tagPath.forEach((tagPart, index) => {
+        if (!current[tagPart]) {
+          const tagId = tagPath.slice(0, index + 1).join('/');
```
Pop0097, line 36: Just want to confirm we won't get an out of bounds index error here

ui/src/components/tag-tree.tsx
```diff
@@ -0,0 +1,103 @@
...
+          isCollapsed={isCollapsed}
+          links={[
+            {
+              title: 'All Tags',
+              icon: Tags,
+              variant: 'ghost',
+            },
+          ]}
+        />
+      </div>
```
Pop0097: **Question:** Would it be useful to lock the "All tags" button in place and only scroll the tree? That way it's easy for the user to find the button. 

https://github.com/user-attachments/assets/9ee0dde5-274c-4471-9df6-a102f017726c

Pop0097 (approved):
Uzair, this is some really good work! All I have are small nits!

One more nit that I have - could the background colour on the selected tag be more visible? I feel like the current shade is a bit difficult to see. 

https://github.com/user-attachments/assets/dcf28804-7f10-473f-b82e-894399f65cf3

ui/src/components/tag-tree.tsx
Uzair5162: We won't since `index` will always be < `tagPath.length` and `.slice` end is exclusive. Even if `index + 1` happened to be out of bounds, `.slice` will just return the slice that is within bounds (ex. `["a","b","c"].slice(0,10)` returns `["a","b","c"]`).

ui/src/components/command-list.tsx
```diff
@@ -9,14 +9,20 @@ import { Command } from '@/types/command';
...
 }
 
-export function CommandList({ items }: CommandListProps) {
+export function CommandList({ items, selectedTag }: CommandListProps) {
   const [command, setCommand] = useCommand();
 
   return (
     <ScrollArea className="h-[calc(100vh-121px)]">
       <div className="flex flex-col gap-2 p-4 pt-0">
+        {selectedTag && (
```
dylan-p-wong: Should this maybe be hoisted out of CommandList? So we don't pass selectedTag as a prop to CommandList.

dylan-p-wong (approved):

ui/src/components/tag-tree.tsx
Uzair5162: I agree, changed it to only scroll the tree.

Uzair5162:
> Uzair, this is some really good work! All I have are small nits!
> 
> One more nit that I have - could the background colour on the selected tag be more visible? I feel like the current shade is a bit difficult to see.
> 
>  Screen.Recording.2025-01-16.at.22.12.58.mov

Agree, made it darker to match the hover/select colour on the command list.

ui/src/components/command-list.tsx
Uzair5162: Done, (this also makes the badge not scroll with the commands which I think is good).

PR #45, opened 2025-01-18
Command display

dylan-p-wong:
image
- Refactor parameters to allow parameter refresh.
- Add parameter refresh on UI.
- Add basic command display as seen above (Uses form to display but form is not activated).
- Add copy command to clipboard.

### DIFF
--- a/logic/src/command.rs
+++ b/logic/src/command.rs
@@ -4,7 +4,7 @@ use fuzzy_matcher::skim::SkimMatcherV2;
 use fuzzy_matcher::FuzzyMatcher;
 use thiserror::Error;
 
-use crate::param::{ParameterError, ParameterHandler};
+use crate::param::{ParameterError, ParameterHandler, SerializableParameter};
 use crate::Logic;
 
 #[derive(Error, Debug)]
@@ -180,10 +180,10 @@ impl Logic {
 
     #[tokio::main]
     /// Handles the generation of parameters for a command
-    pub async fn generate_parameters(&self, command: Command) -> Result<String, ParameterError> {
-        let parameterized_command =
-            ParameterHandler::default().replace_parameters(command.internal_command.command)?;
-        Ok(parameterized_command)
+    pub async fn generate_parameters(&self, command: String) -> Result<(String, Vec<SerializableParameter>, Vec<String>), ParameterError> {
+        let (parameterized_command, parameters, generated_parameters) =
+            ParameterHandler::default().replace_and_get_parameters(command)?;
+        Ok((parameterized_command, parameters, generated_parameters))
     }
 }
 
@@ -530,9 +530,11 @@ mod tests {
         let commands = list_commands_result.unwrap();
         assert!(commands.len() == 1);
 
-        let generated_param_result = logic.generate_parameters(commands.first().unwrap().clone());
+        let generated_param_result = logic.generate_parameters(commands.first().unwrap().internal_command.command.clone());
         assert!(generated_param_result.is_ok());
-        let generated_param = generated_param_result.unwrap();
+        let (generated_param, parameters, generated_parameters) = generated_param_result.unwrap();
         assert_ne!(generated_param, "echo @{int}");
+        assert_eq!(parameters.len(), 1);
+        assert_eq!(generated_parameters.len(), 1);
     }
 }
--- a/logic/src/command.rs
+++ b/logic/src/command.rs
@@ -180,10 +180,14 @@ impl Logic {
 
     #[tokio::main]
     /// Handles the generation of parameters for a command
-    pub async fn generate_parameters(&self, command: String) -> Result<(String, Vec<SerializableParameter>, Vec<String>), ParameterError> {
-        let (parameterized_command, parameters, generated_parameters) =
-            ParameterHandler::default().replace_and_get_parameters(command)?;
-        Ok((parameterized_command, parameters, generated_parameters))
+    pub async fn generate_parameters(&self, command: String) -> Result<(String, Vec<String>), ParameterError> {
+        ParameterHandler::default().replace_parameters(command)
+    }
+
+    #[tokio::main]
+    /// Handles the parsing of parameters for a command
+    pub async fn parse_parameters(&self, command: String) -> Result<(Vec<String>, Vec<SerializableParameter>), ParameterError> {
+        ParameterHandler::default().parse_parameters(command)
     }
 }
 
@@ -532,9 +536,8 @@ mod tests {
 
         let generated_param_result = logic.generate_parameters(commands.first().unwrap().internal_command.command.clone());
         assert!(generated_param_result.is_ok());
-        let (generated_param, parameters, generated_parameters) = generated_param_result.unwrap();
+        let (generated_param,  generated_parameters) = generated_param_result.unwrap();
         assert_ne!(generated_param, "echo @{int}");
-        assert_eq!(parameters.len(), 1);
         assert_eq!(generated_parameters.len(), 1);
     }
 }
--- a/logic/src/param.rs
+++ b/logic/src/param.rs
@@ -89,7 +89,8 @@ impl FromStr for StringParameter {
     type Err = ParameterError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let re = Regex::new(r"string\[(\d+),\s*(\d+)\]").map_err(|_| ParameterError::Parsing)?;
+        let re =
+            Regex::new(r"@\{string\[(\d+),\s*(\d+)\]\}").map_err(|_| ParameterError::Parsing)?;
         if let Some(caps) = re.captures(s) {
             let min = caps[1]
                 .parse::<u32>()
@@ -106,7 +107,7 @@ impl FromStr for StringParameter {
         }
 
         match s {
-            "string" => Ok(StringParameter::default()),
+            "@{string}" => Ok(StringParameter::default()),
             _ => Err(ParameterError::InvalidParameter),
         }
     }
@@ -135,7 +136,7 @@ impl FromStr for IntParameter {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
-            "int" => Ok(IntParameter::default()),
+            "@{int}" => Ok(IntParameter::default()),
             _ => Err(ParameterError::InvalidParameter),
         }
     }
@@ -165,7 +166,7 @@ impl FromStr for BooleanParameter {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
-            "boolean" => Ok(BooleanParameter::default()),
+            "@{boolean}" => Ok(BooleanParameter::default()),
             _ => Err(ParameterError::InvalidParameter),
         }
     }
@@ -187,7 +188,7 @@ impl RandomStringGenerator for BooleanParameter {
 pub enum SerializableParameter {
     Int(IntParameter),
     String(StringParameter),
-    Boolean(BooleanParameter)
+    Boolean(BooleanParameter),
 }
 
 impl RandomStringGenerator for SerializableParameter {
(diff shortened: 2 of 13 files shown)


logic/src/param.rs
```diff
@@ -132,19 +136,20 @@ impl FromStr for IntParameter {
...
 
-impl Parameter for IntParameter {
+impl RandomStringGenerator for IntParameter {
     fn generate_random(&self, rng: &mut dyn RandomNumberGenerator) -> String {
         let random_int = rng.generate_range(self.min, self.max);
         random_int.to_string()
     }
 }
 
+#[derive(Serialize, Deserialize, Debug)]
```
Pop0097: **Nit:** Does the `BooleanParameter` type need a `min` and `max` property?

ui/src-tauri/src/lib.rs
```diff
@@ -81,14 +81,34 @@ fn delete_command(command: DeleteCommand) -> Result<(), String> {
...
+
+#[tauri::command]
+fn replace_parameters(command: String) -> Result<(String, Vec<String>), String> {
+    let logic = Logic::try_default().map_err(|e| format!("Failed to initialize Logic: {:?}", e))?;
+
+    logic
+        .generate_parameters(command)
+        .map_err(|e| format!("Error generating parameters: {:?}", e))
+}
+
```
Pop0097: **Question:** Is it possible to define an error enum using `thiserror::Error` instead of returning a string?

ui/src/components/param.tsx
```diff
@@ -0,0 +1,23 @@
...
+  generatedValue: string;
+}
+
+export function Param({ parameter, index, generatedValue }: ParamProps) {
+  return (
+    <>
+      <div key={index.toString()} className="text-sm">
+        <Label>
+          {parameter.type} (Min: {parameter.data.min.toString()}, Max:{' '}
+          {parameter.data.max.toString()})
```
Pop0097: **Nit:** If we are printing a boolean, showing the min and max field doesn't make much sense. Also, why is there an additional `{' '}` at the end of L16?

Screenshot 2025-01-20 at 07 54 45

ui/src/components/param.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface ParamProps {
+  parameter: Parameter;
+  index: Number;
+  generatedValue: string;
+}
+
+export function Param({ parameter, index, generatedValue }: ParamProps) {
+  return (
+    <>
```
Pop0097: **Nit:** If i recall correctly, the `<>` is not needed since this component is wrapped in a `<div>`

ui/src/components/param.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <div key={index.toString()} className="text-sm">
+        <Label>
+          {parameter.type} (Min: {parameter.data.min.toString()}, Max:{' '}
+          {parameter.data.max.toString()})
+        </Label>
+        <Input value={generatedValue} />
```
Pop0097: **NIt:** If the parameter text is very long, it gets clipped and there's no way for the user to see what's past the boundary of the text box. This could be a non-issue, but wanted to bring it up

Screenshot 2025-01-20 at 07 59 10

Pop0097 (commented):
Great work!

ui/src-tauri/src/lib.rs
dylan-p-wong: I think so. Tauri should be able to handle that. Does something like ```{"error": "..."}``` make sense to you? That is roughly what it would appear as in the UI (before displaying the actual string).

ui/src/components/param.tsx
dylan-p-wong: It is inside a ScrollArea should it should allow scrolling. However we should for sure style it up better in a future PR.

Pop0097: I'm referring to the `<Input>` element specifically, not the overall parameter box

ui/src-tauri/src/lib.rs
Pop0097: We could probably do something like 

```rs
#[derive(Error)]
pub enum ParameterError {
    #[error("Failed to initialize logic")]
    LogicInit(#[from] LogicInitError),
    #[error("Failed to parse parameters")]
    Parse(#[from] LogicParameterError),
    
    ...
}
```

I may have messed up some of the type names, but I was picturing an enum with this form.

Pop0097: And then your methods can look like this:

```rs
#[tauri::command]
fn replace_parameters(command: String) -> Result<(String, Vec<String>), ParameterError > {
    let logic = Logic::try_default()?;

    logic.generate_parameters(command)?
}
```

logic/src/param.rs
dylan-p-wong: Removed.

ui/src/components/param.tsx
dylan-p-wong: Removed the min/max for boolean.

logic/src/command.rs
```diff
@@ -180,10 +180,14 @@ impl Logic {
...
-            ParameterHandler::default().replace_parameters(command.internal_command.command)?;
-        Ok(parameterized_command)
+    pub async fn generate_parameters(&self, command: String) -> Result<(String, Vec<String>), ParameterError> {
+        ParameterHandler::default().replace_parameters(command)
+    }
+
+    #[tokio::main]
+    /// Handles the parsing of parameters for a command
+    pub async fn parse_parameters(&self, command: String) -> Result<(Vec<String>, Vec<SerializableParameter>), ParameterError> {
+        ParameterHandler::default().parse_parameters(command)
```
danyal002, line 190: Why do we need 2 `parse_parameters` methods?

danyal002, line 190: I'll fix this with Dhruv later this week, but `ParameterHandler` needs to abstract away the random number generator somehow to simplify other code.

ui/src-tauri/src/lib.rs
```diff
@@ -1,6 +1,35 @@
...
+    LogicInit(#[from] LogicInitError),
+    #[error("Failed to parse parameters")]
+    Parse(#[from] ParameterError),
+    #[error("Failed to delete command")]
+    DeleteCommand(#[from] DeleteCommandError),
+    #[error("Failed to add command")]
+    AddCommand(#[from] AddCommandError),
+    #[error("Failed to list commands")]
+    ListCommand(#[from] ListCommandError),
+}
```
danyal002: There's a better way to do this, and that's what a big part of our refactor was before. Each command type should have its own enum struct. In the `cli` crate, we have `pub enum HandleAddError` for handling add errors. The reason for that is when you start doing error handling, you have to match on every single type which might be unrelated. For example, if you're in some `Add` context, you're forced to handle errors for `Delete` as well. Refer to `cli.command.{any operation}_command.rs`.

ui/src-tauri/src/lib.rs
```diff
@@ -1,6 +1,35 @@
...
+
+// we must manually implement serde::Serialize (https://github.com/tauri-apps/tauri/discussions/8805)
+impl serde::Serialize for UIError {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        serializer.serialize_str(self.to_string().as_ref())
+    }
+}
```
danyal002, line 34: I can't immediately tell, but why do we need this? Are converting to and from json anywhere or something like that?

danyal002 (changes requested):

ui/src/components/param.tsx
danyal002: If the parameter box is inside a scroll area, then we \*should\* be able to scroll horizontally.

logic/src/command.rs
dylan-p-wong: We don't need it but it makes a more clear pattern of always initializing the logic handler to perform operations.

ui/src-tauri/src/lib.rs
dylan-p-wong: I can change to that but the use case for this error is to serialize in the front-end. So we will never be performing matches against this error. My thought process behind adding in one struct is so we do not have to put ```impl serde::Serialize for UIError {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::ser::Serializer,
    {
        serializer.serialize_str(self.to_string().as_ref())
    }
}``` for every type of error.

ui/src-tauri/src/lib.rs
dylan-p-wong: We are bridging the gap between our rust code and the UI.

ui/src/components/command-display.tsx
```diff
@@ -1,97 +1,253 @@
...
+          <div className="flex items-center p-4">
             <div className="flex items-start gap-4 text-sm">
-              <Avatar>
-                <AvatarImage alt={command.id} />
-                <AvatarFallback>{command.id}</AvatarFallback>
-              </Avatar>
               <div className="grid gap-1">
-                <div className="font-semibold">{command.command}</div>
+                <div className="font-semibold">
+                  {command.tag ? command.tag : 'Untagged'}
```
Uzair5162: are we sure we want the tag up here as the header?

ui/src/components/command-display.tsx
```diff
@@ -1,97 +1,253 @@
...
-            <DropdownMenuItem>Option 3</DropdownMenuItem>
-            <DropdownMenuItem>Option 4</DropdownMenuItem>
-          </DropdownMenuContent>
-        </DropdownMenu>
-      </div>
-      <Separator />
       {command ? (
         <div className="flex flex-1 flex-col">
-          <div className="flex items-start p-4">
+          <div className="flex items-center p-4">
```
Uzair5162: styling nit probably for future PR: need this height to match the other panes

ui/src/components/param.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function Param({ parameter, index, generatedValue }: ParamProps) {
+  return (
+    <div key={index.toString()} className="text-sm">
+      <Label>
+        {parameter.type}{' '}
+        {(parameter.type == ParameterType.String ||
+          parameter.type == ParameterType.Int) &&
+          `Min: ${parameter.data.min.toString()}, Max: ${parameter.data.max.toString()})`}
```
Uzair5162: Missing opening `(`
image

ui/src/components/command-display.tsx
```diff
@@ -1,97 +1,253 @@
...
+                </Button>
+              </TooltipTrigger>
+              <TooltipContent>Edit command</TooltipContent>
+            </Tooltip>
+            <RemoveDialog command={command} />
           </div>
           <Separator />
           <div className="flex-1 whitespace-pre-wrap p-4 text-sm">
-            {command.note}
+            <Form {...form}>
```
Uzair5162: styling nit probably for future PR: this section should be in a `ScrollArea` so the whole window doesn't scroll

ui/src/components/command-display.tsx
dylan-p-wong: I just roughly followed this. 
image

Uzair5162 (approved):

Pop0097 (approved):

ui/src-tauri/src/lib.rs
```diff
@@ -26,27 +57,23 @@ impl From<&Command> for DisplayCommand {
...
+fn list_commands() -> Result<Vec<DisplayCommand>, UIError> {
+    let logic = Logic::try_default()?;
+    let commands = logic.list_commands(false, false)?;
     let commands: Vec<DisplayCommand> = commands.iter().map(DisplayCommand::from).collect();
     Ok(commands)
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct AddCommand {
+pub struct ExternalCommand {
```
danyal002: With all the refactors done on the backend, could you please ensure that we don't already have a type that we can reuse instead of creating a new type? I feel like we have this already.

ui/src-tauri/src/lib.rs
```diff
@@ -57,14 +84,19 @@ impl From<&AddCommand> for InternalCommand {
...
 
-    logic
-        .add_command(internal_command)
-        .map_err(|e| format!("Error adding command: {:?}", e))
+#[tauri::command]
+fn update_command(command_id: i64, command: ExternalCommand) -> Result<(), UIError> {
+    let logic = Logic::try_default()?;
+    let internal_command = InternalCommand::from(&command);
+    logic.update_command(command_id, internal_command)?;
+    Ok(())
```
danyal002: This can be simplified to:

`Ok(logic.update_command(command_id, InternalCommand::from(&command))?)`

ui/src-tauri/src/lib.rs
```diff
@@ -73,12 +105,24 @@ pub struct DeleteCommand {
 }
 
 #[tauri::command]
-fn delete_command(command: DeleteCommand) -> Result<(), String> {
-    let logic = Logic::try_default().map_err(|e| format!("Failed to initialize Logic: {:?}", e))?;
+fn delete_command(command: DeleteCommand) -> Result<(), UIError> {
+    let logic = Logic::try_default()?;
+    logic.delete_command(command.id)?;
+    Ok(())
```
danyal002:
```suggestion
Ok(logic.delete_command(command.id)?)
```

ui/src-tauri/src/lib.rs
```diff
@@ -73,12 +105,24 @@ pub struct DeleteCommand {
...
+}
+
+#[tauri::command]
+fn parse_parameters(command: String) -> Result<(Vec<String>, Vec<SerializableParameter>), UIError> {
+    let logic = Logic::try_default()?;
+    let ret = logic.parse_parameters(command)?;
+    Ok(ret)
```
danyal002: same comment

ui/src-tauri/src/lib.rs
```diff
@@ -73,12 +105,24 @@ pub struct DeleteCommand {
...
+}
 
-    logic
-        .delete_command(command.id)
-        .map_err(|e| format!("Error deleting command: {:?}", e))
+#[tauri::command]
+fn replace_parameters(command: String) -> Result<(String, Vec<String>), UIError> {
+    let logic = Logic::try_default()?;
+    let ret = logic.generate_parameters(command)?;
+    Ok(ret)
```
danyal002: same comment

danyal002 (changes requested):

ui/src-tauri/src/lib.rs
dylan-p-wong: I switched to this but personally I think it makes it harder to read.

ui/src-tauri/src/lib.rs
dylan-p-wong: Now using InternalCommand.

ui/src/components/param.tsx
dylan-p-wong: Will fix later.

ui/src-tauri/src/lib.rs
danyal002: I disagree. It's very idiomatic in rust to do this.

danyal002 (approved):

ui/src-tauri/src/lib.rs
dylan-p-wong: It is because ```logic.update_command(command_id, InternalCommand::from(&command))``` returns void what makes it confusing.

PR #46, opened 2025-01-19
Rework UI logo

Uzair5162:
Replaces the "CmdStack" text in the corner to be the actual logo and makes it collapsible.
image
image


danyal002:
This looks amazing! Awesome job! 

This probably should be another PR, but can we make sure that clicking the logo will always bring the user back to the "home" screen? like let's say if someone was searching and the stack was being filtered with search results, or if some tag was selected and the stack was being filtered by tag. If the user clicks the logo, then everything should be reset.

danyal002 (approved):

Pop0097 (approved):

PR #47, opened 2025-01-25
Add config file for CLI

Pop0097:
```bash
cmdstack config                     # displays wizard for both print style and display limit
cmdstack config -p all           # only displays wizard for display limit
cmdstack config -d 10           # only displays wizard for print style
cmdstack config -p all -d 10 # only displays success message
```

Screenshot 2025-01-25 at 18 00 25


cli/src/config.rs
dylan-p-wong, line 1: Should this stuff be moved to the logic layer? So we can implement settings/config for the UI aswell.

dylan-p-wong: Also if we plan to share this config across the UI should we store in SQLite so it manages all the race conditions and other database stuff for us?

danyal002: > Also if we plan to share this config across the UI should we store in SQLite so it manages all the race conditions and other database stuff for us?

That is a good idea. I didn't think about that. The only nice thing about having a JSON file is that you can open the file and edit the settings yourself. It's akin to giving a surgeon a scalpel to make expert cuts. By having the config inside our database, we remove that ability. A surgeon would be pissed if they had to make a cut with a kitchen knife (going through the UI). I'm honestly not sure what to do here. @Pop0097  @Uzair5162  any thoughts?

Notably, VSCode allows you to modify the JSON directly and provides a UI to change the settings. As for race conditions, I don't think that'll be much of a problem. First, it's highly unlikely that the user will change settings through the CLI whilst also having the settings page open _already_ in the GUI. For the sake of argument, let's assume that they do. What we can do to prevent race conditions is keep a lock on the file while we're modifying stuff. We show the user an error if the file is currently in use.

cli/src/args.rs
```diff
@@ -90,3 +104,15 @@ pub struct ImportExportArgs {
     /// The relative path of the file
     pub file: String,
 }
+
+/// Arguments for setting the CLI config
+#[derive(Debug, Args, Clone)]
+pub struct ConfigArgs {
+    /// Configure how commands are displayed
+    #[clap(long = "print-style", short = 'p', value_enum)]
```
danyal002, line 112: ~~I'd argue that me make `short = 'ps'` for consistency.~~ nvm, I wrote this before I wrote the essay above.

danyal002:
I don't particularly agree with the approach used to make changes to the config. You currently have:
```
cmdstack config                     # displays wizard for both print style and display limit
cmdstack config -p all           # only displays wizard for display limit
cmdstack config -d 10           # only displays wizard for print style
```

That's a bit unintuitive. Here's why: when a user wants the change the config of something, they most likely know exactly what they want to change. The main subject of the `config` command is some particular setting, so using a flag like `-p` doesn't make sense. I would replace that with `cmdstack config print-style {new_value}`, so the structure would be `cmdstack config {what_configuration} {new_value}`. (perhaps we should rethink how we do command updates too🤔)

In the `cmdstack config --help`, we can display the settings that a user is allowed to change along with acceptable values/constraints and defaults. I don't think we need to use any interactive CLI here either.

Another comment I'd make is that we implement this in a similar way to how it exists in VSCode or GitHub CLI. They tend to group configs with periods. For example, `git config --global --add --bool push.autoSetupRemote true`, emphasis on the `push.autoSetupRemote true` part. Right now, we don't have a lot of settings, but soon, we're going to have plenty, especially when we start implementing keyboard shortcuts. So, `display.lineLimit` could be the new config key for line limits. `print-style` should also be renamed to something else. On its own, it doesn't really tell a new user what it means. A better config key could be `display.numLines` or something. You can ping me to discuss this further.

Later, we should provide more granularity and differ between `ui` and `cli` config keys.

danyal002 (changes requested):

PR #48, opened 2025-01-30
Massive refactor & config command implementation

danyal002:
This PR was created in collaboration with @Pop0097. 

### Changes:
* Implemented the `cmdstack config` command based on @danyal002's recommendations on @Pop0097's previous PR that is now closed. 
* Removed the `logic::init` calls from various places. I created a `Cli` struct where `logic` is a property. It is initialized once when the program starts up and is reused everywhere else. This simplified a lot of error handling. 
*  Incorporated the new `config` logic to get `print_style` and other properties from the `config` that is now part of the `Cli` struct. All configs are in one central place now.
* Split the `import_export` files into separate files and added them to the `commands` module for consistency. 
* Various renaming of methods/structs and formatting for better readability.

### Follow up changes after this PR gets merged:
- Rename `cli.command` module to `handlers`
- Remove `{}_command` from all handlers sub-modules.
- Consolidate the `AddArgs` and `UpdateArgs` into one since they’re the same.


cli/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+            let val: u32 = config_args
+                .value
+                .parse()
+                .map_err(|_| ConfigError::InvalidValue)?;
+            config.cli_display_limit = val;
+        }
+        _ => {
+            return Err(ConfigError::UnknownProperty);
+        }
+    }
```
Pop0097: Need to add a handler for `cli_display_by_most_recently_used`. Alternatively, we should remove `cli_display_by_most_recently_used` from the `Config` struct in the logic crate

Pop0097 (commented):
Overall lgtm

cli/src/command/search_utils.rs
```diff
@@ -33,176 +33,130 @@ pub fn check_search_args_exist(command: &Option<String>, tag: &Option<String>) -
...
+            &format_output(
+                &("<bold>Select a command</bold> <italics>".to_owned()
+                    + columns
+                    + "</italics><bold>:</bold>"),
+            ),
+            formatted_commands,
+        )
+        // Only display the command once the user makes a selection
+        .with_formatter(&|i| commands[i.index].internal_command.command.to_string())
+        .with_page_size(self.config.cli_display_limit as usize)
```
Pop0097, line 89: https://github.com/user-attachments/assets/df6df74c-0eb2-488b-b786-8112d180dcb9

This does not appear to be working as intended.

cli/src/command/search_utils.rs
```diff
@@ -33,176 +33,130 @@ pub fn check_search_args_exist(command: &Option<String>, tag: &Option<String>) -
...
-            Cell::new(if command.internal_command.favourite {
-                "*"
-            } else {
-                ""
-            }),
-        ]));
+    /// Formats the commands for printing based on the user's preferred style.
+    /// Returns the columns to be printed
+    fn format_commands_for_printing(&self, commands: &Vec<Command>) -> (Vec<String>, &str) {
+        match self.config.cli_print_style {
```
Pop0097, line 104: https://github.com/user-attachments/assets/4d5c17dd-702d-42ed-9a02-2515a4cf7cbd


Also this

cli/src/config.rs
danyal002: Thanks, I forgot to remove it from the `Config` struct after confirming with you.

cli/src/command/search_utils.rs
danyal002: Uhh this is not erroring for me for some reason.

cli/src/command/search_utils.rs
danyal002: ~~This is broken for me too. Looking...~~ Fixed

cli/src/command/search_utils.rs
danyal002: Fixed. I was reading the defaults instead of the file.

cli/src/args.rs
```diff
@@ -90,3 +86,12 @@ pub struct ImportExportArgs {
     /// The relative path of the file
     pub file: String,
 }
+
+/// Arguments for setting the CLI config
+#[derive(Debug, Args, Validate)]
+pub struct ConfigArgs {
+    pub property: String,
```
dylan-p-wong: Should this also have ```#[validate(length(min = 1))]```?

cli/src/command/delete_command.rs
```diff
@@ -1,56 +1,54 @@
...
-        .map_err(|e| match e {
-            FetchSearchCandidatesError::NoCommandsFound => HandleDeleteError::NoCommandFound,
-            _ => HandleDeleteError::SearchCandidates(e),
-        })?;
+        let search_results = self
+            .logic
+            .search_command(SearchCommandArgs {
+                command: search_user_input.command,
+                tag: search_user_input.tag,
+                order_by_use: args.recent,
```
dylan-p-wong: Maybe rename recent->order_by_use?

cli/src/command/search_utils.rs
```diff
@@ -33,176 +33,130 @@ pub fn check_search_args_exist(command: &Option<String>, tag: &Option<String>) -
...
-}
-
 #[derive(Error, Debug)]
 pub enum PromptUserForCommandSelectionError {
     #[error("No commands found")]
-    NoCommandsFound,
-    #[error("failed to get selected item")]
-    InquireError(#[from] InquireError),
+    NoCommandsProvided,
+    #[error("failed to render")]
```
dylan-p-wong: This error message seems misleading.

logic/src/config.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    pub fn write(&self) -> Result<(), ConfigError> {
+        let config_path = Config::config_file_path()?;
+        let config_content = serde_json::to_string_pretty(self).map_err(ConfigError::Serialize)?;
+        fs::write(config_path, config_content)?;
+        Ok(())
+    }
+
+    fn config_file_path() -> Result<PathBuf, ConfigError> {
+        let mut config_dir = dirs::config_dir().ok_or_else(|| {
```
dylan-p-wong: We can probably consolidate this with the other logic to get the sqlite path.

logic/src/config.rs
```diff
@@ -0,0 +1,87 @@
+use serde::{Deserialize, Serialize};
+use std::fs;
+use std::path::PathBuf;
+use std::str::FromStr;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum ConfigError {
+    #[error("Invalid value provided")]
+    InvalidValue,
```
dylan-p-wong: Could parameterize this error. "Invalid value provided: {}"

cli/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+    #[error("Invalid value provided")]
+    InvalidValue,
+    #[error("Failed to read config")]
+    ReadConfig(#[source] logic::config::ConfigError),
+    #[error("Failed to read config")]
+    WriteConfig(#[source] logic::config::ConfigError),
+}
+
+pub fn handle_config_command(config_args: ConfigArgs) -> Result<(), ConfigError> {
+    let mut config = logic::config::Config::read().map_err(ConfigError::ReadConfig)?;
```
dylan-p-wong: I would rather specify ```ConfigError::ReadConfig``` where the error is sourced.

cli/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+                .parse()
+                .map_err(|_| ConfigError::InvalidValue)?;
+            config.cli_display_limit = val;
+        }
+        _ => {
+            return Err(ConfigError::UnknownProperty);
+        }
+    }
+
+    config.write().map_err(ConfigError::WriteConfig)
```
dylan-p-wong: I would rather specify ```ConfigError::WriteConfig``` where the error is sourced.

cli/src/command/update_command.rs
```diff
@@ -71,34 +70,34 @@ pub fn set_command_properties_wizard(
...
-        })?;
+        let search_results = self
+            .logic
+            .search_command(SearchCommandArgs {
+                command: search_user_input.command,
+                tag: search_user_input.tag,
+                order_by_use: args.recent,
+                favourites_only: args.favourite,
+            })
+            .map_err(HandleUpdateError::Search)?;
```
dylan-p-wong: Can this be removed? I thought that is the point of the error handling stuff.

cli/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+    ReadConfig(#[source] logic::config::ConfigError),
+    #[error("Failed to read config")]
+    WriteConfig(#[source] logic::config::ConfigError),
+}
+
+pub fn handle_config_command(config_args: ConfigArgs) -> Result<(), ConfigError> {
+    let mut config = logic::config::Config::read().map_err(ConfigError::ReadConfig)?;
+
+    match config_args.property.as_ref() {
+        "cli_print_style" => {
```
Uzair5162: could we have these config properties displayed on `cmdstack config` instead of erroring? similarly, on something like `cmdstack config cli_print_style` with no value we could show value options.

cli/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[derive(Debug, Error)]
+pub enum ConfigError {
+    #[error("Unknown property provided")]
+    UnknownProperty,
+    #[error("Invalid value provided")]
+    InvalidValue,
+    #[error("Failed to read config")]
+    ReadConfig(#[source] logic::config::ConfigError),
+    #[error("Failed to read config")]
```
Uzair5162: should say `write`

Uzair5162 (changes requested):

cli/src/args.rs
danyal002: I added it, but it's not necessary. `serde` will throw a serialization error because it couldn't serialized a string to a property in the `Config` struct.

cli/src/command/search_utils.rs
danyal002: Okay, 2 questions:
* Why?
* What would be a better error message?

logic/src/config.rs
danyal002: I don't think I understand what you're trying to say. Do you mean that we can get both paths from the same place? Or standardize how we get the paths? Or store the config along with where we store the sqlite db file?

cli/src/config.rs
danyal002: Can you rephrase what you're trying to say please?

cli/src/command/update_command.rs
danyal002: Good catch. Thanks!

cli/src/config.rs
danyal002: ~~Hmm, same question as above. @Pop0097 can you take a look please?~~

cli/src/config.rs
danyal002: Dude, that's a great idea. Also, you just fixed a problem that we were having yesterday. We couldn’t figure out a better way to read/write to the JSON whilst also taking values from the users. After googling for how to display the possible values for `--help`, I figured out how we can implement this in a 10x better way. Updates to follow.

danyal002: Reworked how `config` works to stick to standards. You need to type `--help` for it to work, but it will show you possible values.
EDIT: I figured it out! There's a derive macro that we can use that will print help if no arg is provided.

Uzair5162 (approved):
A lot easier to use now with config hints 👍🏽

dylan-p-wong (approved):

Pop0097 (approved):

cli/src/command/search_utils.rs
dylan-p-wong: I think I read the code wrong.

dylan-p-wong (approved):

cli/src/command/config_command.rs
```diff
@@ -0,0 +1,81 @@
...
+pub enum CliPrintStyle {
+    #[default]
+    All,
+    CommandsOnly,
+}
+
+impl Cli {
+    /// Handles the config modification command
+    pub fn handle_config_command(&self, config_args: ConfigArgs) -> Result<(), ConfigError> {
+        let mut config = logic::config::Config::read().map_err(ConfigError::ReadConfig)?;
```
dylan-p-wong: can this map_err be removed?

cli/src/command/config_command.rs
```diff
@@ -0,0 +1,81 @@
...
+            }
+            ConfigArgs::CliDisplayLimit(cli_display_limit_args) => {
+                cli_display_limit_args
+                    .validate()
+                    .map_err(|e| ConfigError::InvalidValue(e.to_string()))?;
+
+                config.cli_display_limit = cli_display_limit_args.value
+            }
+        }
+        config.write().map_err(ConfigError::WriteConfig)
```
dylan-p-wong: can this map_err be removed?

cli/src/command/config_command.rs
danyal002: Yep, forgot to do that.

PR #52, opened 2025-02-03
style header, scroll content, styling

Uzair5162:
- styles the command view header to be the same height as the other headers
- makes the command view content scroll instead of the whole window scrolling, with the generated command fixed at the bottom
- reduces spacing between param refresh button and text
- other small styling things

before:
image
after:
image


danyal002:
Looks nice! 
* I'm not sure if the generated parameters box being fixed at the bottom looks good or not. I think if we calibrate the default window size of the app based on percentage of the screen, create default sizes for the Note (with the option of dragging to make it bigger), we wouldn't have the need to fix the generated parameters box. You'd be able to see it directly especially since the Parameters section is scrollable. Thoughts?

dylan-p-wong (approved):

danyal002 (approved):

PR #55, opened 2025-02-05
Refactored the way parameter generation is implemented for easier maintenance

danyal002:
This PR was co-authored with @Pop0097. Changelog:
* Refactored how parameter generation works by decoupling the parsing and parameter generation. 
* Added 4 new configs to the config file for string and int parameter lengths and ranges respectively.
* Implemented ability to change those config values via the CLI
* Fixed a UI bug where default parameters were not being fetched from the config. Uses one instance of the logic crate protected by a RWLock and added that to the Tauri state.
* Improvements on how we capture the parameters from a command string. We use optional grouping in the regex pattern now. 

Please clone the branch and test the cli and ui thoroughly and comment if there's any issues.


cli/src/command/config_command.rs
```diff
@@ -50,6 +56,28 @@ pub enum CliPrintStyle {
...
+
+#[derive(Debug, Args, Validate)]
+#[command(arg_required_else_help(true))]
+pub struct ParamIntArgs {
+    /// The minimum value for integer parameters
+    pub min: i32,
+
+    /// The maximum value for integer parameters
+    pub max: i32,
+}
```
Pop0097: It seems like we requiring the user to update both properties at the same time

Screenshot 2025-02-05 at 18 49 57

Screenshot 2025-02-05 at 18 52 24

I think we should allow the user to update both independently. Perhaps we need to do the following:

```rs
// In ConfigArgs
{
    /// Modify string parameter min limits
     ParamStringMinLength(ParamStringArg),

    /// Modify string parameter max limits
     ParamStringMaxLength(ParamStringArg),

     /// Modify string parameter min limits
     ParamIntMinRange(ParamIntArg),

    /// Modify string parameter max limits
     ParamIntMinRange(ParamIntArg),
}


[derive(Debug, Args, Validate)]
 #[command(arg_required_else_help(true))]
 pub struct ParamStringArgs{
     #[validate(range(min = 1, message = "The value must be at least 1"))]
     pub value: u32,
 }

 #[derive(Debug, Args, Validate)]
 #[command(arg_required_else_help(true))]
 pub struct ParamIntArg {
     pub value: i32,
 }
```

logic/src/command.rs
```diff
@@ -4,7 +4,8 @@ use fuzzy_matcher::skim::SkimMatcherV2;
 use fuzzy_matcher::FuzzyMatcher;
 use thiserror::Error;
 
-use crate::param::{ParameterError, ParameterHandler, SerializableParameter};
+// use crate::param::{ParameterError, ParameterHandler, SerializableParameter};
```
Pop0097:
```suggestion
```

logic/src/command.rs
```diff
@@ -67,7 +68,7 @@ impl Logic {
             return Err(AddCommandError::InvalidInput);
         }
 
-        ParameterHandler::default().validate_parameters(command.command.clone())?;
+        self.parse_parameters(command.command.clone())?;
```
Pop0097, line 70: Just a thought when we improve error handling and logging, currently if the provided parameters are invalid, we only display the following: 

Screenshot 2025-02-05 at 18 59 43

We should ideally come up with better error messages to display to the user.

logic/src/lib.rs
```diff
@@ -6,27 +6,30 @@ pub mod command;
 pub mod config;
 pub mod import_export;
 pub mod param;
+pub mod parameters;
```
Pop0097, line 8: **nit:** Should remove `pub mod param`

logic/src/parameters/boolean.rs
```diff
@@ -0,0 +1,28 @@
+use serde::{Deserialize, Serialize};
+use std::str::FromStr;
+
+use super::{populator::RandomNumberGenerator, GenerateRandomValues, ParameterError};
+
+#[derive(Serialize, Deserialize, Debug, Default)]
```
Pop0097: IntParameter and StringParameter don't derive Default. I confirmed this builds

```suggestion
#[derive(Serialize, Deserialize, Debug)]
```

logic/src/parameters/parser.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        for mat in re.find_iter(&command) {
+            match self.parse_parameter(mat.as_str().to_owned()) {
+                Ok(param) => {
+                    parameters.push(param);
+                }
+                Err(e) => {
+                    return Err(e);
+                }
+            }
```
Pop0097:
```suggestion
            let param = self.parse_parameter(mat.as_str().to_owned())?;
            parameters.push(param);
```

logic/src/parameters/parser.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        // There should be a parameter for each "gap" between strings
+        assert_eq!(non_parameter_strs.len() - 1, parameters.len());
+
+        Ok((non_parameter_strs, parameters))
+    }
+
+    fn parse_parameter(&self, s: String) -> Result<SerializableParameter, ParameterError> {
+        if let Ok(ph) = StringParameter::from_str(&s, &self.config) {
```
Pop0097: **Nit:** What does `ph` stand for? Maybe we can call them `string_param`, `int_param`, `bool_param` and so on

logic/src/parameters/populator.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+impl Logic {
+    pub fn populate_parameters(
+        &self,
+        non_parameter_strs: Vec<String>,
+        parameters: Vec<SerializableParameter>,
+    ) -> Result<(String, Vec<String>), ParameterError> {
+        let mut rng = rand::rngs::ThreadRng::default();
```
Pop0097: If we want to allow for dependency injection, we can make this method take in an option parameter of type `but dyn RandomNumberGenerator`. If the parameter is none, we can use `rand::rigs::ThreadRng`.

ui/src-tauri/src/lib.rs
```diff
@@ -62,23 +71,35 @@ impl From<&Command> for DisplayCommand {
 }
 
 #[tauri::command]
-fn list_commands() -> Result<Vec<DisplayCommand>, UIError> {
-    let logic = Logic::try_default()?;
-    let commands = logic.list_commands(false, false)?;
-    let commands: Vec<DisplayCommand> = commands.iter().map(DisplayCommand::from).collect();
-    Ok(commands)
+fn list_commands(state: State<Ui>) -> Result<Vec<DisplayCommand>, UIError> {
+    if let Ok(logic) = state.logic.write() {
```
Pop0097, line 75: https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.write

`RwLock::write()` blocks the current thread. Could this result in our UI becoming unresponsive?

cli/src/command/config_command.rs
danyal002: I had that before. The reason why I changed it was validation (min <= max), but I fixed that by having that if statement. I'm not sure what's a good idea here. It's cumbersome if you want to change both at the same time because you have to run two commands. Maybe a better way would be to do something like `cmdstack config param-string-length --min 2 --max 5` and make both optional. Thoughts?

logic/src/command.rs
danyal002: Yeah I noticed that too in the UI. That should be fixed once we start doing the logging. Added new task to task list.

logic/src/parameters/parser.rs
danyal002: I have no idea. I don't remember writing that though :P

ui/src-tauri/src/lib.rs
danyal002: I'm not sure. I'll test it out. However, I got that example from Tauri's documentation: https://docs.rs/tauri/latest/tauri/trait.Manager.html#method.manage

cli/src/command/config_command.rs
dylan-p-wong: I wouldn't even add config for these values imo.

Pop0097: >  Maybe a better way would be to do something like cmdstack config param-string-length --min 2 --max 5 and make both optional. Thoughts?

I like this idea

ui/src-tauri/src/lib.rs
```diff
@@ -1,15 +1,22 @@
...
     },
-    param::{ParameterError, SerializableParameter},
+    parameters::{parser::SerializableParameter, ParameterError},
     Logic, LogicInitError,
 };
 use serde::{Deserialize, Serialize};
+use tauri::State;
 use thiserror::Error;
 
+pub struct Ui {
```
dylan-p-wong, line 16: nit: rename UI or change the other "UI" -> "Ui"

logic/src/param.rs
```diff
@@ -1,389 +0,0 @@
...
-}
-
-impl Default for ParameterHandler {
-    fn default() -> Self {
-        let rng = rand::thread_rng();
-        Self { rng: Box::new(rng) }
-    }
-}
-
-#[cfg(test)]
```
dylan-p-wong, line 297: All these tests are gone?

dylan-p-wong (changes requested):

logic/src/parameters/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+pub mod parser;
+pub mod populator;
+pub mod string;
+
+pub trait FromStrWithConfig: Sized {
+    fn from_str(s: &str, config: &Config) -> Result<Self, ParameterError>;
+}
+
+pub trait GenerateRandomValues {
+    fn generate_random_values(&self, rng: &mut dyn RandomNumberGenerator) -> String;
```
Uzair5162: nit: rename to be singular (`generate_random_value()`) since this is called on individual params

logic/src/parameters/int.rs
```diff
@@ -0,0 +1,63 @@
...
+
+            return Ok(Self { min, max });
+        }
+        Err(ParameterError::InvalidParameter)
+    }
+}
+
+impl GenerateRandomValues for IntParameter {
+    fn generate_random_values(&self, rng: &mut dyn RandomNumberGenerator) -> String {
+        rng.generate_range(self.min, self.max + 1).to_string()
```
Uzair5162: this should just be `generate_range(self.min, self.max)`. `ThreadRng`'s impl of `generate_range()` accepts inclusive bounds:

https://github.com/danyal002/cmd-stack/blob/18b8a5d073d7e381507b7577e5aa926777b9b0ce/logic/src/parameters/populator.rs#L12

logic/src/parameters/int.rs
```diff
@@ -0,0 +1,63 @@
...
+            } else {
+                config.param_int_max
+            };
+
+            if min > max {
+                return Err(ParameterError::InvalidMinMax(
+                    min.to_string(),
+                    max.to_string(),
+                ));
+            }
```
Uzair5162: nit: should consider putting this into a function if we plan on supporting more types with min/max since its pretty much duplicated from the `StringParameter.from_str()`

Uzair5162 (changes requested):

logic/src/parameters/populator.rs
danyal002: There's only one call to this method, and I don't see why we'd need to pass in a randomizer like that. If we have another `rng` later, we can do it then.

ui/src-tauri/src/lib.rs
danyal002: I didn't look too much into this, but when I was testing, it was working fine. Also, if the UI becomes unresponsive during a write, there's something drastically wrong because the right shouldn't be taking that long.

ui/src-tauri/src/lib.rs
danyal002: I think the Rust convention is `Ui` (PascalCase), so I'll change that.

logic/src/parameters/int.rs
danyal002: fixed that and improved it further by using `(low..=high)`

logic/src/parameters/int.rs
danyal002: Yes, we did that and had to delete it. The reason was that we're also providing default values for the parameters in the else clauses for the min and max. Those types are different for string and int (`u32` and `i32` respectively). Therefore, you can't have one method even though the logic looks the same. If we end up needing something similar, we can extract one of these out (probably the string one) and then reuse it.

logic/src/param.rs
danyal002: I will let @Pop0097 take care of this.

logic/src/parameters/populator.rs
Uzair5162: There's only one call to this with one `rng` because you removed all the tests that were calling this with `self.rng` being a `MockRng` instead of a `ThreadRng`. See:

https://github.com/danyal002/cmd-stack/blob/30a4b998f7a0436c0ff062a12efceb479a899bcf/logic/src/param.rs#L228-L249 and https://github.com/danyal002/cmd-stack/blob/30a4b998f7a0436c0ff062a12efceb479a899bcf/logic/src/param.rs#L312

Uzair5162 (approved):

dylan-p-wong (approved):

PR #57, opened 2025-02-13
Style command display header and tag improvements

Uzair5162:
- styles the selected command's tag in the display pane's header
- changes the last used timestamp to "Last used ... ago" in the display pane's header with a tooltip showing the timestamp (without milliseconds)
- untagged commands in the command list have an outlined tag instead of the usual
- icon beside the selected tag being filtered on
- sorts tags alphabetically at every level of the tree in the tag pane

Before:
image

After:
image

### DIFF
--- a/ui/src/components/command-display.tsx
+++ b/ui/src/components/command-display.tsx
@@ -27,6 +27,9 @@ import { Label } from './ui/label';
 import { Textarea } from './ui/textarea';
 import { Tooltip, TooltipContent, TooltipTrigger } from './ui/tooltip';
 import { ScrollArea } from './ui/scroll-area';
+import { Badge } from './ui/badge';
+import { cn } from '@/lib/utils';
+import { formatDistanceToNow } from 'date-fns';
 
 interface CommandDisplayProps {
   command: Command | null;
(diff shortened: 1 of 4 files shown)


danyal002:
That looks awesome! Great job!

A couple of comments:
* In the `After` screenshot, in the tag pane on the left, `databases` seems to be selected. However, in the command display view, at the top, the `postgres` tag seems to be highlighted. Is that on purpose?
* Do you think we should just remove the highlight in the `postgres` tag in the command view? I think there's enough indicators (2, first in the tag pane, second in the search) to show which tag the command is located under.

Uzair5162:
> That looks awesome! Great job!
> 
> A couple of comments:
> 
> * In the `After` screenshot, in the tag pane on the left, `databases` seems to be selected. However, in the command display view, at the top, the `postgres` tag seems to be highlighted. Is that on purpose?
> * Do you think we should just remove the highlight in the `postgres` tag in the command view? I think there's enough indicators (2, first in the tag pane, second in the search) to show which tag the command is located under.

- yes, inspired by postman again to just highlight the leaf tag of the current command regardless of the current filter:
image

- the tag pane and search just show whats being filtered on (in the screenshots we're filtering on `docker/databases`, which could have multiple leaf tags, not just `postgres`). So the only indicator of the command's tag itself in the tag in the command card in the list, so I think its fine.

danyal002 (approved):

PR #59, opened 2025-02-17
Implemented feature to execute command in shell upon selection by the user

danyal002:
This PR adds user functionality to edit your generated command in case you need to apply final touches and then allows the user to select whethere they want to copy it to their clipboard or execute in their terminal. I made the conscious decision here to not support cmd.exe or PowerShell. We need to investigate further whether that's worth the effort. `.exec()` only works for unix-based OS'. Windows does not let you take over a process like Linux does. If you guys have any suggestions, feel free to comment.

Demo for command execution:

https://github.com/user-attachments/assets/b4fa21d0-a0fa-4d9c-a5da-e652f87e6362

### DIFF
--- a/cli/src/handlers/search.rs
+++ b/cli/src/handlers/search.rs
@@ -4,6 +4,7 @@ use crate::{
         check_search_args_exist, copy_to_clipboard, CopyTextError,
         PromptUserForCommandSelectionError, SearchArgsUserInput,
     },
+    outputs::spacing,
     Cli,
 };
 use inquire::InquireError;
@@ -52,6 +53,8 @@ impl Cli {
         let (text_to_copy, _) = self
             .logic
             .generate_parameters(user_selection.internal_command.command)?;
+
+        spacing();
 
         // Prompt the user to edit the generated command
         let user_edited_cmd = self.prompt_user_for_command_edit(&text_to_copy)?;
--- a/cli/src/handlers/search.rs
+++ b/cli/src/handlers/search.rs
@@ -9,6 +9,7 @@ use crate::{
 use inquire::InquireError;
 use log::error;
 use logic::command::{SearchCommandArgs, SearchCommandError};
+use std::{os::unix::process::CommandExt, process::Command};
 use thiserror::Error;
 
 #[derive(Error, Debug)]
@@ -52,7 +53,22 @@ impl Cli {
             .logic
             .generate_parameters(user_selection.internal_command.command)?;
 
-        copy_to_clipboard(text_to_copy)?;
+        // Prompt the user to edit the generated command
+        let user_edited_cmd = self.prompt_user_for_command_edit(&text_to_copy)?;
+
+        // Prompt the user for command action
+        let action = self.prompt_user_for_action()?;
+
+        if action == "Execute" {
+            // Note: using `.exec()` will shutdown our app and execute the command, therefore, we can't handle errors.
+            let _ = self.logic.update_command_last_used_prop(user_selection.id);
+            let _ = Command::new("sh")
+                .arg("-c")
+                .arg(user_edited_cmd.clone())
+                .exec();
+        }
+
+        copy_to_clipboard(user_edited_cmd)?;
 
         Ok(self
             .logic
--- a/cli/src/handlers/cli_prompter.rs
+++ b/cli/src/handlers/cli_prompter.rs
@@ -71,7 +71,7 @@ impl Cli {
         initial_value: &str,
     ) -> Result<String, InquireError> {
         Text::new(&format_output(
-            "<bold>Edit</bold> (Press enter to continue): ",
+            "<bold>Edit</bold> <italics>(Press enter to continue)</italics><bold>:</bold> ",
         ))
         .with_initial_value(initial_value)
         .prompt()
--- a/cli/src/handlers/cli_prompter.rs
+++ b/cli/src/handlers/cli_prompter.rs
@@ -65,6 +65,28 @@ impl Cli {
         })
     }
 
+    /// Prompt user to edit the generated command
+    pub fn prompt_user_for_command_edit(
+        &self,
+        initial_value: &str,
+    ) -> Result<String, InquireError> {
+        Text::new(&format_output(
+            "<bold>Edit</bold> (Press enter to continue): ",
+        ))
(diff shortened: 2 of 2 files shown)


cli/src/command/search_command.rs
```diff
@@ -52,7 +53,22 @@ impl Cli {
...
+        if action == "Execute" {
+            // Note: using `.exec()` will shutdown our app and execute the command, therefore, we can't handle errors.
+            let _ = self.logic.update_command_last_used_prop(user_selection.id);
+            let _ = Command::new("sh")
+                .arg("-c")
+                .arg(user_edited_cmd.clone())
+                .exec();
+        }
+
+        copy_to_clipboard(user_edited_cmd)?;
```
Pop0097:
```suggestion
        if action == "Execute" {
            // Note: using `.exec()` will shutdown our app and execute the command, therefore, we can't handle errors.
            let _ = self.logic.update_command_last_used_prop(user_selection.id);
            let _ = Command::new("sh")
                .arg("-c")
                .arg(user_edited_cmd.clone())
                .exec();
        } else {
            copy_to_clipboard(user_edited_cmd)?;
        }
```

cli/src/command/search_command.rs
```diff
@@ -52,7 +53,22 @@ impl Cli {
             .logic
             .generate_parameters(user_selection.internal_command.command)?;
 
-        copy_to_clipboard(text_to_copy)?;
+        // Prompt the user to edit the generated command
+        let user_edited_cmd = self.prompt_user_for_command_edit(&text_to_copy)?;
```
Pop0097: I think we may need to provide more context to users here so that they know to hit ENTER if they want to keep the original command (or edit the text if they want to change it).

cli/src/command/search_command.rs
danyal002: Don't need an `else` clause because program terminates at `.exec()` if inside the if-statement.

cli/src/command/search_command.rs
danyal002: I'll add some text to the prompt.

danyal002: Fixed.

cli/src/command/search_utils.rs
```diff
@@ -65,6 +65,28 @@ impl Cli {
         })
     }
 
+    /// Prompt user to edit the generated command
+    pub fn prompt_user_for_command_edit(
+        &self,
+        initial_value: &str,
+    ) -> Result<String, InquireError> {
+        Text::new(&format_output(
+            "<bold>Edit</bold> (Press enter to continue): ",
```
Pop0097: For consistency: 

```suggestion
            "<bold>Edit</bold> <italics>(Press enter to continue)</italics><bold>:</bold> ",
```

cli/src/command/search_command.rs
```diff
@@ -52,7 +53,22 @@ impl Cli {
...
+        // Prompt the user for command action
+        let action = self.prompt_user_for_action()?;
+
+        if action == "Execute" {
+            // Note: using `.exec()` will shutdown our app and execute the command, therefore, we can't handle errors.
+            let _ = self.logic.update_command_last_used_prop(user_selection.id);
+            let _ = Command::new("sh")
+                .arg("-c")
+                .arg(user_edited_cmd.clone())
+                .exec();
```
Pop0097: Screenshot 2025-02-17 at 20 57 15

Would it look better if we had an empty line between the "Select a command" and "Edit" prompts?

Pop0097 (commented):
In a future PR, we should probably create some flags that the user can set to skip the edit and/or action prompts. 

Perhaps we could also create configs (`cli-show-edit-on-select` and `cli-action-on-select`) that the user can customize to have a default action (ex. "skip edit by default", "copy by default", and "Execute by default")

cli/src/command/search_command.rs
danyal002: I would honestly prefer if we somehow cleared the `Select a command` part. I'll look into it or just add a space.

Pop0097 (approved):

cli/src/handlers/search.rs
```diff
@@ -52,7 +54,24 @@ impl Cli {
...
+        // Prompt the user for command action
+        let action = self.prompt_user_for_action()?;
+
+        if action == "Execute" {
+            // Note: using `.exec()` will shutdown our app and execute the command, therefore, we can't handle errors.
+            let _ = self.logic.update_command_last_used_prop(user_selection.id);
+            let _ = Command::new("sh")
+                .arg("-c")
+                .arg(user_edited_cmd.clone())
+                .exec();
```
Uzair5162, line 71: Would `.output()` instead of `.exec()` be better?

danyal002: Not really tbh. Output will spawn a child process and wait for the command to finish. We don’t want anything to do with that. If a user starts docker or something, we're cooked.

Uzair5162 (approved):

cli/src/handlers/search.rs
```diff
@@ -52,7 +54,24 @@ impl Cli {
...
+
+        // Prompt the user to edit the generated command
+        let user_edited_cmd = self.prompt_user_for_command_edit(&text_to_copy)?;
+
+        // Prompt the user for command action
+        let action = self.prompt_user_for_action()?;
+
+        if action == "Execute" {
+            // Note: using `.exec()` will shutdown our app and execute the command, therefore, we can't handle errors.
+            let _ = self.logic.update_command_last_used_prop(user_selection.id);
```
dylan-p-wong, line 67: Should we handle the error on ```self.logic.update_command_last_used_prop(user_selection.id);``` ?

dylan-p-wong (approved):

cli/src/handlers/search.rs
danyal002: Refer to my other comment for the details, but basically, the program terminates on the call to execute(), and everything after is dead code.

dylan-p-wong: But this is before the ```exec```? For example changing ```let _ = self.logic.update_command_last_used_prop(user_selection.id);``` to ``` self.logic.update_command_last_used_prop(user_selection.id)?;```

PR #62, opened 2025-02-19
Open command in terminal and generated command styling

Uzair5162:
moves generated cmd textarea to the top and adds buttons inside it to copy to clipboard and open the command in terminal
- the generated command can be edited in place before copying/opening in terminal
  - I think opening in terminal will only work for mac, it opens a terminal session (or uses an existing one) and types in the command without executing
  - This will require giving the app permission to type
- the text area expands to be multi-line for long commands
- the commands `last_used` is updated when either button is clicked

Before:
image

After:
image

https://github.com/user-attachments/assets/c429beef-69fc-49c9-9aaa-77c997c059c9

### DIFF
--- a/ui/src-tauri/src/lib.rs
+++ b/ui/src-tauri/src/lib.rs
@@ -177,6 +177,30 @@ fn write_config(config: Config, state: State<Ui>) -> Result<(), UiError> {
     Err(UiError::Race)
 }
 
+#[tauri::command]
+fn open_terminal(command: String) {
+    std::process::Command::new("osascript")
+        .arg("-e")
+        .arg(format!(
+            "tell application \"Terminal\" to activate"
+        ))
+        .arg("-e")
+        .arg(format!(
+            "tell application \"System Events\" to keystroke \"{}\"",
+            command
+        ))
+        .spawn()
+        .unwrap();
+}
+
+#[tauri::command]
+fn update_command_last_used(command_id: i64, state: State<Ui>) -> Result<(), UiError> {
+    if let Ok(logic) = state.logic.write() {
+        return Ok(logic.update_command_last_used_prop(command_id)?);
+    }
+    Err(UiError::Race)
+}
+
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
     let logic = Logic::try_default()
@@ -197,7 +221,9 @@ pub fn run() {
             update_command,
             search_commands,
             read_config,
-            write_config
+            write_config,
+            update_command_last_used,
+            open_terminal
         ])
         .run(tauri::generate_context!())
         .expect("error while running tauri application");
--- a/ui/src-tauri/src/lib.rs
+++ b/ui/src-tauri/src/lib.rs
@@ -181,9 +181,7 @@ fn write_config(config: Config, state: State<Ui>) -> Result<(), UiError> {
 fn open_terminal(command: String) {
     std::process::Command::new("osascript")
         .arg("-e")
-        .arg(format!(
-            "tell application \"Terminal\" to activate"
-        ))
+        .arg(format!("tell application \"Terminal\" to activate"))
         .arg("-e")
         .arg(format!(
             "tell application \"System Events\" to keystroke \"{}\"",
--- a/ui/src-tauri/src/lib.rs
+++ b/ui/src-tauri/src/lib.rs
@@ -38,6 +38,8 @@ pub enum UiError {
     ConfigWriteError(#[from] ConfigWriteError),
     #[error("Failed to obtain lock to complete the required action")]
     Race,
+    #[error("Failed to execute command in terminal")]
+    ExecuteCommand,
 }
 
 // we must manually implement serde::Serialize (https://github.com/tauri-apps/tauri/discussions/8805)
@@ -178,17 +180,16 @@ fn write_config(config: Config, state: State<Ui>) -> Result<(), UiError> {
 }
 
 #[tauri::command]
-fn open_terminal(command: String) {
+fn execute_in_terminal(command: String) -> Result<(), UiError> {
     std::process::Command::new("osascript")
         .arg("-e")
-        .arg(format!("tell application \"Terminal\" to activate"))
-        .arg("-e")
         .arg(format!(
-            "tell application \"System Events\" to keystroke \"{}\"",
+            "tell application \"Terminal\" to activate do script \"{}\" in window 1",
             command
         ))
         .spawn()
-        .unwrap();
+        .map_err(|_| UiError::ExecuteCommand)?;
+    Ok(())
 }
 
 #[tauri::command]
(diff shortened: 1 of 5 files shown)


ui/src/components/command-display.tsx
```diff
@@ -127,12 +133,25 @@ export function CommandDisplay({ command }: CommandDisplayProps) {
...
   }
 
-  function onCopy(e: { preventDefault: () => void }) {
-    e.preventDefault();
+  function onCopy() {
     navigator.clipboard.writeText(generatedCommand);
     toast({
       title: 'Copied to clipboard ✅',
     });
+
```
dylan-p-wong, line 152: I think we should setup the error handling here. I think we should toast on ```open_terminal``` errors but leave silent on ```update_command_last_used``` and just console.log.

ui/src/components/command-display.tsx
```diff
@@ -222,7 +241,52 @@ export function CommandDisplay({ command }: CommandDisplayProps) {
               </div>
               <Separator />
               <div className="flex flex-1 flex-col">
-                <ScrollArea className="h-[calc(100vh-180px)]">
+                <div className="p-4">
```
dylan-p-wong, line 258: Can we bring this into its own component?

ui/src/components/ui/textarea.tsx
```diff
@@ -10,15 +10,15 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
     return (
       <textarea
         className={cn(
-          "flex min-h-[60px] w-full rounded-md border border-input bg-transparent px-3 py-2 text-sm shadow-sm placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50",
```
dylan-p-wong, line 13: Just make sure this does not mess with any other places we use Textarea.

ui/src-tauri/src/lib.rs
```diff
@@ -177,6 +177,28 @@ fn write_config(config: Config, state: State<Ui>) -> Result<(), UiError> {
...
+    std::process::Command::new("osascript")
+        .arg("-e")
+        .arg(format!("tell application \"Terminal\" to activate"))
+        .arg("-e")
+        .arg(format!(
+            "tell application \"System Events\" to keystroke \"{}\"",
+            command
+        ))
+        .spawn()
+        .unwrap();
```
dylan-p-wong: Should we handle this error rather than just unwrapping?

danyal002 (changes requested):
I took a quick look at your PR (on my phone) for the new generated command box. It looks really nice. I have a few comments:

* I don’t know if it’s a good idea to have the open in terminal and copy button side by side. I’ve personally been a victim of accidentally pressing the other. 

* for opening in terminal, we want it to be executed, not typed. The reason for this is that we’re limited by the APIs of the platforms. It might work on Mac, but not on linux or windows. The other thing is that you’ll need keyboard/input permissions which is a big no no.  That’s asking for permissions we don’t need. 

* can you post a screenshot of what it looks like for a multi line command?

* I don’t think the layout of a generated command and then the saved command stacked on top of each other looks good or not. Could you prototype moving that where it was before?

ui/src/components/command-display.tsx
Uzair5162: done

ui/src/components/command-display.tsx
Uzair5162: i'll do this in the pr styling the other stuff

ui/src/components/ui/textarea.tsx
Uzair5162: good point, undid it

ui/src-tauri/src/lib.rs
Uzair5162: done

Uzair5162:
> I took a quick look at your PR (on my phone) for the new generated command box. It looks really nice. I have a few comments:
> 
> * I don’t know if it’s a good idea to have the open in terminal and copy button side by side. I’ve personally been a victim of accidentally pressing the other.
> * for opening in terminal, we want it to be executed, not typed. The reason for this is that we’re limited by the APIs of the platforms. It might work on Mac, but not on linux or windows. The other thing is that you’ll need keyboard/input permissions which is a big no no.  That’s asking for permissions we don’t need.
> * can you post a screenshot of what it looks like for a multi line command?
> * I don’t think the layout of a generated command and then the saved command stacked on top of each other looks good or not. Could you prototype moving that where it was before?

- Leaving buttons as is as we discussed (can revisit later)
- Changed it to execute in terminal without needing extra permissions
- Made it so that it grows up to 3 lines and then scrolls
- The rest of the panel hasn't been styled yet

image

dylan-p-wong:
I feel the parameter regeneration is too disconnected from the command actions. I think they should be closer or maybe a regenerate parameter button in the text box.

danyal002:
> I feel the parameter regeneration is too disconnected from the command actions. I think they should be closer or maybe a regenerate parameter button in the text box.

Yeah, I agree. The layout doesn't feel fluid at all. Like we have the generated command at the top, then we have details of the command and then a box with parameters in it. Let's discuss this more on discord.

ui/src/components/command-display.tsx
```diff
@@ -127,12 +133,36 @@ export function CommandDisplay({ command }: CommandDisplayProps) {
     }
   }
 
-  function onCopy(e: { preventDefault: () => void }) {
-    e.preventDefault();
+  function onUseCommand() {
+    invoke('update_command_last_used', {
+      commandId: command?.id,
+    }).catch((error) => {
+      console.error(error);
```
danyal002, line 140: We're not showing this to the user? Perhaps a toaster like `An error occurred whilst updating metadata. Please refer to logs.` would be helpful I think.

danyal002:
One other issue I found is that executing in terminal will open the `Terminal` app instead of the default terminal app on Mac. 

https://github.com/user-attachments/assets/c9aefceb-98e9-4390-a0a8-2390a95ab99a

Like for me, `iTerm2` is the default. We'll need to come back to this or fix it here. I think this problem exists in the CLI as well.

ui/src/components/command-display.tsx
Uzair5162: done

Uzair5162:
> One other issue I found is that executing in terminal will open the `Terminal` app instead of the default terminal app on Mac.
> 
>  Screen.Recording.2025-02-23.at.12.15.28.mov 
> Like for me, `iTerm2` is the default. We'll need to come back to this or fix it here. I think this problem exists in the CLI as well.

Not sure how to make this happen let's come back to it

danyal002 (approved):

dylan-p-wong (approved):

PR #64, opened 2025-02-21
[CLI] Blank Params

Pop0097:
Add blank parameters to the CLI. Made in collaboration with @danyal002.

Here is a demo of the command in action: 

https://github.com/user-attachments/assets/366cffbd-056b-431b-97ec-1305ec0465b2

### DIFF
--- a/cli/src/handlers/cli_prompter.rs
+++ b/cli/src/handlers/cli_prompter.rs
@@ -112,7 +112,7 @@ impl Cli {
             format_output(
                 &self
                     .logic
-                    .number_parameters(&commands[i.index].internal_command.command),
+                    .index_parameters_for_display(&commands[i.index].internal_command.command),
             )
         })
         .with_page_size(self.logic.config.cli_display_limit as usize)
@@ -199,6 +199,7 @@ impl Cli {
                 _ => None,
             })
             .collect::<Result<_, _>>()?;
+        spacing();
         Ok(blank_param_values)
     }
 }
--- a/cli/src/handlers/cli_prompter.rs
+++ b/cli/src/handlers/cli_prompter.rs
@@ -9,6 +9,7 @@ use data::models::Command;
 use inquire::{InquireError, Select, Text};
 use log::error;
 use prettytable::{format, Cell, Row, Table};
+use regex::Regex;
 use termion::terminal_size;
 use thiserror::Error;
 
@@ -71,7 +72,7 @@ impl Cli {
         initial_value: &str,
     ) -> Result<String, InquireError> {
         Text::new(&format_output(
-            "<bold>Edit</bold> <italics>(Press enter to continue)</italics><bold>:</bold> ",
+            "<bold>Edit Command</bold> <italics>(Press enter to continue)</italics><bold>:</bold> ",
         ))
         .with_initial_value(initial_value)
         .prompt()
@@ -80,7 +81,7 @@ impl Cli {
     /// Prompt user to select an action: Copy or Execute
     pub fn prompt_user_for_action(&self) -> Result<String, InquireError> {
         Ok(Select::new(
-            &format_output("<bold>Action:</bold>"),
+            &format_output("<bold>Select Action:</bold>"),
             vec!["Copy", "Execute"],
         )
         .prompt()?
@@ -107,7 +108,11 @@ impl Cli {
             formatted_commands,
         )
         // Only display the command once the user makes a selection
-        .with_formatter(&|i| commands[i.index].internal_command.command.to_string())
+        .with_formatter(&|i| {
+            format_output(
+                &self.get_selected_command_string(&commands[i.index].internal_command.command),
+            )
+        })
         .with_page_size(self.logic.config.cli_display_limit as usize)
         .raw_prompt()
         {
@@ -118,6 +123,21 @@ impl Cli {
         };
 
         Ok(commands[selected_command.index].clone())
+    }
+
+    fn get_selected_command_string(&self, command: &str) -> String {
+        let blank_param_regex = Regex::new(r"@\{\s*\}").unwrap();
+        let mut blank_param_num = 1;
+
+        let command_with_placeholders =
+            blank_param_regex.replace_all(command, |_: &regex::Captures| {
+                let replacement =
+                    format!("<bold><italics>@{{{}}}</italics></bold>", blank_param_num);
+                blank_param_num += 1;
+                replacement
+            });
+
+        command_with_placeholders.to_string()
     }
 
     /// Formats the commands for printing based on the user's preferred style.
--- a/cli/src/handlers/cli_prompter.rs
+++ b/cli/src/handlers/cli_prompter.rs
@@ -183,7 +183,7 @@ impl Cli {
 
     pub fn fill_blank_params(
         &self,
-        parsed_params: &Vec<SerializableParameter>,
+        parsed_params: &[SerializableParameter],
     ) -> Result<Vec<String>, PromptUserForCommandSelectionError> {
         let blank_param_values: Vec<String> = parsed_params
             .iter()
--- a/cli/src/handlers/cli_prompter.rs
+++ b/cli/src/handlers/cli_prompter.rs
@@ -110,7 +110,7 @@ impl Cli {
         // Only display the command once the user makes a selection
         .with_formatter(&|i| {
             format_output(
-                &self.get_selected_command_string(&commands[i.index].internal_command.command),
+                &self.format_selected_command(&commands[i.index].internal_command.command),
             )
         })
         .with_page_size(self.logic.config.cli_display_limit as usize)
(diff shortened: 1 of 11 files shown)


cli/src/handlers/cli_prompter.rs
```diff
@@ -120,6 +125,22 @@ impl Cli {
         Ok(commands[selected_command.index].clone())
     }
 
+    /// Numbers blank parameters in the selected command
```
dylan-p-wong: Can we move to the logic crate and change the name to something like ```number_blank_params```

logic/src/parameters/populator.rs
```diff
@@ -18,6 +18,7 @@ impl Logic {
         &self,
         non_parameter_strs: Vec<String>,
         parameters: Vec<SerializableParameter>,
+        blank_parameter_values: Vec<String>,
```
dylan-p-wong, line 21: I wonder if removing ```blank_parameter_values``` and adding a ```Optional<String>``` value field to ```BlankParameter``` and then just calling ```generate_random_value``` (or renaming to ```generate_value```) would clean up the logic. No strong opinion just something to consider.

logic/src/parameters/populator.rs
```diff
@@ -167,11 +237,30 @@ mod tests {
...
+    fn test_populate_parameters_blank_missing() {
+        let logic = Logic::try_default().unwrap();
+
+        let rng = Box::new(MockRng::new(vec![2, 0]));
+
+        let non_parameter_strs = vec!["ls ".to_string(), "".to_string()];
+        let parameters = vec![SerializableParameter::Blank];
+        let blank_params_values = vec![];
+
+        let ret = logic.populate_parameters(
```
dylan-p-wong, line 282: Lets also add a test case for ```blank_params_values.len > number_of_blank_parameters```.

logic/src/parameters/populator.rs
```diff
@@ -141,6 +162,55 @@ mod tests {
...
+            " ".to_string(),
+            " ".to_string(),
+            " ".to_string(),
+            "".to_string(),
+        ];
+        let parameters = vec![
+            SerializableParameter::Blank,
+            SerializableParameter::Int(IntParameter::default()),
+            SerializableParameter::Blank,
+            SerializableParameter::Boolean(BooleanParameter {}),
```
dylan-p-wong: We should create ```BooleanParameter::default()``` if it's not too much effort.

cli/src/handlers/cli_prompter.rs
```diff
@@ -107,7 +108,11 @@ impl Cli {
...
         // Only display the command once the user makes a selection
-        .with_formatter(&|i| commands[i.index].internal_command.command.to_string())
+        .with_formatter(&|i| {
+            format_output(
+                &self.format_selected_command(&commands[i.index].internal_command.command),
+            )
+        })
         .with_page_size(self.logic.config.cli_display_limit as usize)
         .raw_prompt()
         {
```
danyal002: Should use the `?` operator here for error handling.

cli/src/handlers/cli_prompter.rs
```diff
@@ -120,6 +125,22 @@ impl Cli {
         Ok(commands[selected_command.index].clone())
     }
 
+    /// Numbers blank parameters in the selected command
+    ///
+    /// ex. 'git commit \"@{} @{}\"' becomes 'git commit \"@{1} @{2}\"'
+    fn format_selected_command(&self, command: &str) -> String {
```
danyal002: This method could be named differently. `format` is very generic. Perhaps something like `index_selected_command_parameters`

cli/src/handlers/search.rs
```diff
@@ -49,12 +52,41 @@ impl Cli {
...
-        let (text_to_copy, _) = self
+        let (non_param_strings, parsed_params) = self
             .logic
-            .generate_parameters(user_selection.internal_command.command)?;
+            .parse_parameters(user_selection.internal_command.command.clone())?;
+
+        let has_blank_params = parsed_params
+            .iter()
+            .any(|item| matches!(item, SerializableParameter::Blank));
+        let blank_param_values = if has_blank_params {
```
danyal002, line 62: You can use a `filter_map` here like such:
```
let blank_param_values: Vec<String> = parsed_params
    .iter()
    .enumerate()
    .filter_map(|(index, param)| match param {
        SerializableParameter::Blank => {
            if index == 0 {
                Output::BlankParameter.print();
            }
            let prompt_text = format!("<bold>Fill in @{{{}}}:</bold>", index + 1);
            Some(Text::new(&format_output(&prompt_text)).prompt())
        }
        _ => None,
    })
    .collect::<Result<_, _>>()?;
```

cli/src/handlers/search.rs
```diff
@@ -49,12 +52,41 @@ impl Cli {
...
+            Output::BlankParameter.print();
+            let mut blank_param_num = 1;
+            // Prompt the user once for each blank parameter in the command
+            let values = parsed_params
+                .iter()
+                .filter_map(|param| match param {
+                    SerializableParameter::Blank => {
+                        let prompt_text = format!("<bold>Fill in @{{{}}}:</bold>", blank_param_num);
```
danyal002: Hmm, you're breaking abstraction here. The search handler should not have code that's printing stuff. All that should go in the `cli_prompter`

logic/src/parameters/populator.rs
danyal002: Yeah, I agree. The user-provided blank-parameter value will show up in the struct itself and can be unwrapped, otherwise, it can be an empty vector.

ui/src-tauri/src/lib.rs
```diff
@@ -132,7 +132,7 @@ fn parse_parameters(
 #[tauri::command]
 fn replace_parameters(command: String, state: State<Ui>) -> Result<(String, Vec<String>), UiError> {
     if let Ok(logic) = state.logic.write() {
-        return Ok(logic.generate_parameters(command)?);
+        return Ok(logic.generate_parameters(command, vec![])?);
```
danyal002: Can we avoid using macros? They're generally not recommended because they make code harder to read. An alternative would be `Vec::new()` or something.

danyal002 (changes requested):

logic/src/parameters/populator.rs
```diff
@@ -29,14 +30,34 @@ impl Logic {
...
+                            let user_val = blank_parameter_values[blank_param_used_index].clone();
+                            blank_param_used_index += 1;
+                            user_val
+                        } else {
+                            let total_blank_params_needed = parameters
+                                .iter()
+                                .filter(|p| matches!(p, SerializableParameter::Blank))
+                                .count();
+                            return Err(ParameterError::MissingBlankParamValues(
+                                (blank_param_used_index + 1).to_string(),
```
Uzair5162: Shouldn't have a `+ 1` here. If 0 blank params are provided then `blank_param_used_index` will be 0 and we should error with `0 values provided...`

Uzair5162 (changes requested):

cli/src/handlers/search.rs
```diff
@@ -49,12 +52,26 @@ impl Cli {
...
+        let (non_param_strings, parsed_params) = self
             .logic
-            .generate_parameters(user_selection.internal_command.command)?;
+            .parse_parameters(user_selection.internal_command.command.clone())?;
+
+        let has_blank_params = parsed_params
+            .iter()
+            .any(|item| matches!(item, SerializableParameter::Blank));
+        let blank_param_values = if has_blank_params {
+            self.fill_blank_params(&parsed_params)?
```
Uzair5162, line 63: should be a `spacing();` after this (I think you accidentally removed it in 33509457c5a2baf37f748f7793a9bfeb64ce414e)

image

logic/src/command.rs
```diff
@@ -178,14 +179,30 @@ impl Logic {
...
     ) -> Result<(String, Vec<String>), ParameterError> {
         let (non_parameter_strs, parameters) = self.parse_parameters(command)?;
-        self.populate_parameters(non_parameter_strs, parameters, None)
+        self.populate_parameters(non_parameter_strs, parameters, blank_param_values, None)
+    }
+
+    /// Numbers blank parameters in the selected command
+    ///
+    /// ex. 'git commit \"@{} @{}\"' becomes 'git commit \"@{1} @{2}\"'
+    pub fn number_parameters(&self, command: &str) -> String {
```
danyal002: I thought about this further, and I think a better name for this would be `index_parameters_for_display`. It clearly tells us what this method does and who for.

logic/src/command.rs
```diff
@@ -532,8 +549,10 @@ mod tests {
         let commands = list_commands_result.unwrap();
         assert!(commands.len() == 1);
 
-        let generated_param_result =
-            logic.generate_parameters(commands.first().unwrap().internal_command.command.clone());
+        let generated_param_result = logic.generate_parameters(
+            commands.first().unwrap().internal_command.command.clone(),
+            vec![],
```
danyal002: nit: use `Vec::new()` instead. It's clearer.

danyal002 (approved):

Uzair5162 (approved):

dylan-p-wong (approved):

PR #67, opened 2025-03-05
Command display overhaul

Uzair5162:
Mainly a command display rework, with related changes

#### Blank param indexing:
- refactors `logic::index_parameters_for_display()` to only index blank params in a command, and not style them with `<bold><italics>` tags
  - styling now happens in the cli crate (as it should) and allows for re-use of `logic::index_parameters_for_display()` by the ui
- the _indexed_ command is displayed under the generated command box
image

#### Saved and edit command:
- the saved command is displayed under the generated command box (see above image) with indexed params
- on edit, this turns editable and indexes disappear (showing the original command)
image

#### Note:
- note is displayed below the saved command, with "+ Add a note" button if none
  - on click, the button turns into the text field for the note
image
image

#### Param viewer
- styles the param box and labels blank params with their index
image

#### Misc.
- un-hardcodes the scrollable section height on the viewer (everything below saved command)
- separates some things into components in command-display folder
- commented tag and favourite edit for now, need to be styled in future work

#### Before
image

#### After
image


danyal002:
A few comments:

* Why did you disable editing of parameters? We've been in agreement since before that we will have that box editable.  The reason for this was that it could be the case that the user likes what was randomly generated, but doesn't want all of it, but only part of it. As always, if the user needs to use notepad or another text editor to fix what we generate, then we've failed. Please enable that again.


* Honestly, I don't like how the parameters section looks now. There's just too much going on. I'm not a big fan of the font. It's hard to read (see first screenshot). I would prefer that we revert to the previous implementation where the String min, max was a label on the textfield itself. It looked a lot cleaner. In the current implementation, I'm having to trace my eyes from the String min, max on the left to the field itself on the right.
Screenshot_2025-03-05_at_15 18 07
* I like the "Fill in" placeholder in the blank parameter textfield. Upon reverting to the previous implementation, we can reuse the Blank @{1} text to be the label.

* I don't like how we've removed the Note text box and its heading. It looks like it's not editable when it is. Again, I think we should switch back to the previous implementation.

* I think the order can be changed. The note can be moved all the way down. We can dynamically calculate the height of the parameters scrollpane so that the note section doesn't disappear from view.

* I think there should be more padding between the centre divider line and where the Note and Parameters text are (see second screenshot).
Screenshot_2025-03-05_at_15 19 45

Uzair5162:
> A few comments:
> 
> * Why did you disable editing of parameters? We've been in agreement since before that we will have that box editable.  The reason for this was that it could be the case that the user likes what was randomly generated, but doesn't want all of it, but only part of it. As always, if the user needs to use notepad or another text editor to fix what we generate, then we've failed. Please enable that again.

Parameters were never editable.

> * Honestly, I don't like how the parameters section looks now. There's just too much going on. I'm not a big fan of the font. It's hard to read (see first screenshot). I would prefer that we revert to the previous implementation where the String min, max was a label on the textfield itself. It looked a lot cleaner. In the current implementation, I'm having to trace my eyes from the String min, max on the left to the field itself on the right.
>   Screenshot_2025-03-05_at_15 18 07
> * I like the "Fill in" placeholder in the blank parameter textfield. Upon reverting to the previous implementation, we can reuse the Blank @{1} text to be the label.

I'll change this and generated box font if you have other mono font suggestions. Re: being a label on the textfield itself, I don't like how everything was stacked vertically in the prev ui. I think it looks better side-by-side since it looks less empty. We can discuss alternatives with horizontal spacing and other ppls thoughts.

> * I don't like how we've removed the Note text box and its heading. It looks like it's not editable when it is. Again, I think we should switch back to the previous implementation.

Not sure what you mean. The note box was never editable even though we displayed it in a textarea. It only became editable after clicking edit and that is still the case. Can also make it so that the cursor focuses on the textbox on edit if that helps.

> * I think the order can be changed. The note can be moved all the way down. We can dynamically calculate the height of the parameters scrollpane so that the note section doesn't disappear from view.

Agree.

> * I think there should be more padding between the centre divider line and where the Note and Parameters text are (see second screenshot).
>   Screenshot_2025-03-05_at_15 19 45

Will see how this looks.

Pop0097:
> Honestly, I don't like how the parameters section looks now. There's just too much going on. I'm not a big fan of the font. It's hard to read (see first screenshot). I would prefer that we revert to the previous implementation where the String min, max was a label on the textfield itself. It looked a lot cleaner. In the current implementation, I'm having to trace my eyes from the String min, max on the left to the field itself on the right.

I personally like the new layout better, but I do agree that the font could be changed.

Pop0097:
Some other comments:
- I think when editing a command it isn't very obvious that the user should be clicking the following to change the command (see first image below). I am not too sure what the best path forward for this would be. One idea that I have is removing the command preview box when editing and converting the command to a text box similar to the note (see second image below):

image

image

- When parameters become very long, they get truncated 

Screenshot 2025-03-07 at 13 10 43

Another nit:
- Is there a way that a user can cancel editing? When editing I only see "Save" and "Delete Command" as an option

Uzair5162:
- changed font in params and display/saved command boxes to roboto mono
- hides generated command box on edit
- adds Note heading and moved it below params
- brings param label and display/input closer together and makes overflow scroll
image

Pop0097 (approved):

dylan-p-wong (approved):

ui/src-tauri/src/lib.rs
```diff
@@ -155,6 +156,19 @@ fn replace_parameters(
     Err(UiError::Race)
 }
 
+#[tauri::command]
+fn index_blank_parameters(command: String, state: State<Ui>) -> Result<String, UiError> {
+    if let Ok(logic) = state.logic.write() {
```
danyal002: You might be able to get away with using a `read` lock here since you're not modifying anything.

danyal002 (approved):

PR #71, opened 2025-03-07
Increase the default window size

Pop0097:
New window size: 

Screenshot 2025-03-07 at 15 55 36

### DIFF
--- a/ui/src-tauri/tauri.conf.json
+++ b/ui/src-tauri/tauri.conf.json
@@ -13,9 +13,9 @@
     "windows": [
       {
         "title": "CmdStack",
-        "width": 1000,
-        "height": 600,
-        "minWidth": 800,
+        "width": 1100,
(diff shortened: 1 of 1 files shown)


ui/src-tauri/tauri.conf.json
```diff
@@ -13,9 +13,9 @@
     "windows": [
       {
         "title": "CmdStack",
-        "width": 1000,
-        "height": 600,
-        "minWidth": 800,
+        "width": 1100,
+        "height": 750,
```
danyal002: I would change these to 1440 x 810. This is 75% of the standard 1920x1080 resolution, and since we're releasing on Mac only for now, this would be better as all Macs have very high resolution.

danyal002 (changes requested):

danyal002 (approved):

PR #72, opened 2025-03-07
Fixed logging to enable error chaining

danyal002:
- **Fixed logging in cli crate**
- **Fixed logging in dal and logic crates**
- **Fixed when log file is not accessbile**
- **stuff**
- **Removed the testing code**
- **Fix todo for logic init error handling**

### DIFF
--- a/logic/src/import_export.rs
+++ b/logic/src/import_export.rs
@@ -19,11 +19,11 @@ struct ImportExportFormat {
 
 #[derive(Error, Debug)]
 pub enum ExportError {
-    #[error("Failed to serialize commands")]
+    #[error("Failed to serialize commands: {0}")]
     Deserialize(#[from] serde_json::Error),
-    #[error("Failed to write commands to file")]
+    #[error("Failed to write commands to file: {0}")]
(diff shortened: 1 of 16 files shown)


logic/src/import_export.rs
```diff
@@ -19,21 +19,21 @@ struct ImportExportFormat {
...
 pub enum ImportError {
     #[error("Failed to deserialize commands: {0}")]
     Serialize(#[from] serde_json::Error),
-    #[error("Failed to read commands from file")]
+    #[error("Failed to read commands from file: {0}")]
     Read(String),
-    #[error("Failed to insert commands to the database")]
+    #[error("Failed to insert commands to the database: {0}")]
     Database(#[from] InsertCommandError),
     #[error("File not found at specified path")]
```
Pop0097: **Nit:** I think we should provide more context (ex. what is the file path that was specified)

danyal002: I checked and the last 2 enums weren't even being used apparently.

Pop0097 (approved):

PR #73, opened 2025-03-10
Cancel edit

dylan-p-wong:


Pop0097 (commented):
Great work! I was thinking if we should display an alert when the user clicks "cancel" confirming that they want to cancel editing

danyal002:
Looks great! Can you disable the generated command field when the edit button is clicked? It makes more sense.
Screenshot 2025-03-10 at 16 46 23

dylan-p-wong:
> Looks great! Can you disable the generated command field when the edit button is clicked? It makes more sense. Screenshot 2025-03-10 at 16 46 23

Will do after I pull in the changes from https://github.com/danyal002/cmd-stack/pull/67.

dylan-p-wong:
> Great work! I was thinking if we should display an alert when the user clicks "cancel" confirming that they want to cancel editing

Should we also have a dialog for confirming a save? I am leaning towards having neither.

Uzair5162:
> Looks great! Can you disable the generated command field when the edit button is clicked? It makes more sense. Screenshot 2025-03-10 at 16 46 23

I did this in https://github.com/danyal002/cmd-stack/pull/67

danyal002:
> > Great work! I was thinking if we should display an alert when the user clicks "cancel" confirming that they want to cancel editing
> 
> Should we also have a dialog for confirming a save? I am leaning towards having neither.

Eh I don't think we need it tbh

Uzair5162 (approved):

danyal002 (approved):

dylan-p-wong:
> > Looks great! Can you disable the generated command field when the edit button is clicked? It makes more sense. Screenshot 2025-03-10 at 16 46 23
> 
> Will do after I pull in the changes from #67.

Not doing anymore because of the UI changes no longer need this.

PR #74, opened 2025-03-11
Fixed the execute-in-terminal arguments bug in cli and added functionality to choose which terminal to execute command in for the GUI

danyal002:
This PR does a few things:
1. Previously, the execute-in-terminal for the cli had a bug where not all command arguments would be provided. The output was also not printing. This is now fixed.
2. Added a new config key ui-default-terminal to allow the user to pick their default terminal app to execute commands in from the ui.
3. Added iTerm2 as an option to execute commands in. Default option is still Terminal, but can be easily changed to iTerm using the `cmdstack config ui-default-terminal iterm` command.

### DIFF
--- a/cli/src/handlers/config.rs
+++ b/cli/src/handlers/config.rs
@@ -32,6 +32,9 @@ pub enum ConfigArgs {
 
     /// Modify application theme
     Theme(ApplicationThemeArgs),
+
+    /// Modify default terminal used in the UI to execute commands in
+    UiDefaultTerminal(UiDefaultTerminalArgs),
 }
 
 #[derive(Debug, Args)]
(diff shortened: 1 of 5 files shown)


Pop0097:
Looks good to me so far, but I believe there is some UI work that needs to be done before we can merge this.

Uzair5162:
Changing the default terminal using CLI doesn't seem to reflect in the tauri for me. To reproduce:
- add this println to `ui/src-tauri/src/lib.rs` (since I don't have iTerm)
```#[tauri::command]
fn execute_in_terminal(command: String, state: State<Ui>) -> Result<(), UiError> {
    let mut cmd = std::process::Command::new("osascript");
    if let Ok(logic) = state.logic.read() {
        println!("Using default terminal: {:?}", logic.config.default_terminal); // add this

        match logic.config.default_terminal {
        ...
```

- execute in terminal using the UI button and observe the output
- toggle the default terminal using the cli
- execute in terminal using the UI button and observe the output doesn't change

danyal002:
> * execute in terminal using the UI button and observe the output
> * toggle the default terminal using the cli

This is a known issue. I tested that. You need to restart the UI for it to take effect. Currently, the config is only being read once the UI/CLI starts up. This isn’t a problem with the CLI because the CLI quits after performing one action. It reads the config again once you start `cmdstack` again. It's a problem on the UI though. A fix would be to re-read the config every time we invoke `execute_in_terminal`. If you guys think that's a good enough fix, I can do that.

Uzair5162 (approved):

PR #75, opened 2025-03-12
UI edit tag

Uzair5162:
untagged:
image

---
adding tag:
image

---
editing tag:
image

### DIFF
--- a/ui/src/components/command-display/command-display.tsx
+++ b/ui/src/components/command-display/command-display.tsx
@@ -420,7 +420,7 @@ export function CommandDisplay({ command }: CommandDisplayProps) {
                               </FormControl>
                             ) : (
                               <div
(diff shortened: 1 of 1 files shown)


danyal002 (changes requested):
Great job! That looks awesome! 
Minor nit: can you try lining up the `+Add` text?
Screenshot 2025-03-12 at 11 41 24

danyal002:
Also, can you make it so that when either of the `+Add note` or `+Add tag` is clicked, the appropriate field is focused?

Uzair5162:
> Also, can you make it so that when either of the `+Add note` or `+Add tag` is clicked, the appropriate field is focused?

In a future PR having a hard time with the refs rn

danyal002 (approved):

Pop0097 (approved):

PR #77, opened 2025-03-13
Edit improvements

Uzair5162:
- clicking "+ Add a tag" or "+ Add a note" will only make those editable and focus on their input
- clicking the edit icon makes everything editable and focuses on the command box

### DIFF
--- a/ui/src/components/command-display/command-display.tsx
+++ b/ui/src/components/command-display/command-display.tsx
@@ -10,7 +10,7 @@ import { invoke } from '@tauri-apps/api/core';
 import { formatDistanceToNow } from 'date-fns';
 import format from 'date-fns/format';
 import { Pencil, RefreshCwIcon, Save, X } from 'lucide-react';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useForm } from 'react-hook-form';
 import { z } from 'zod';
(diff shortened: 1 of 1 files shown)


Pop0097:
https://github.com/user-attachments/assets/345d26bc-8e01-4dec-9d0d-bc372a35ba30

Playing around with the UI, it appears that if the user clicks "+ Add a tag" and then "+ Add a note", the focus does not shift. However, in the opposite case, it does shift

ui/src/components/command-display/command-display.tsx
```diff
@@ -174,21 +185,44 @@ export function CommandDisplay({ command }: CommandDisplayProps) {
...
 
+  // This effect handles focusing the correct input when we edit anything
+  useEffect(() => {
+    if (editing.command) {
+      commandRef.current!.focus();
+    } else if (editing.tag) {
+      tagRef.current!.focus();
+    } else if (editing.note) {
+      noteRef.current!.focus();
+    }
```
Pop0097: I think this code is what is causing the focusing behaviour I mentioned in my previous comment

Pop0097:
https://github.com/user-attachments/assets/f31fb9ae-4026-42f4-95c8-6a25400a7a7d

Works great!

Pop0097 (approved):

danyal002 (approved):
Awesome job!

PR #78, opened 2025-03-13
Add default terminal setting to UI

Uzair5162:


Pop0097 (approved):
Nice work. I feel like if we add more settings in the future we will have to consider having some sort of constants file where we store strings such as "Setting updated" etc.

danyal002 (approved):

